PR #1, opened 2015-01-13
Proper boot order

rjschwei:
- regionServiceClient
  - Fix the ordering of the initialization. The registration service needs
    to run prior to cloud-final, if it exists to ensure repositories in on
    demand images are available if a user script is to be executed in the
    cloud-final stage


schaefi:
looks good to me, tested with systemd only though, but the sysV init script looks also clean. Thus merging in :)

PR #2, opened 2015-02-02
Boot order

rjschwei:
- bump 6.3.8
- set service type to oneshot to ensure systemd waits until this unit
  is complete prior to starting wahtever follows.

### DIFF
--- a/regionServiceClient/cloud-regionsrv-client.spec
+++ b/regionServiceClient/cloud-regionsrv-client.spec
@@ -17,7 +17,7 @@
 #
 
 Name:           cloud-regionsrv-client
-Version:        6.3.7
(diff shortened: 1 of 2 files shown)


rjschwei:
As I thought, a simple service does not show as running after exit:

# systemctl status guestregister.service

guestregister.service - Obtain Cloud SMT info and register with the SMT server
   Loaded: loaded (/usr/lib/systemd/system/guestregister.service; enabled)
   Active: inactive (dead) since Mon 2015-02-02 20:01:39 UTC; 10min ago

Thus there is no change in behavior with the current change to "oneshot" also the output does show that something happened.

I think we are OK with this implementation.

PR #5, opened 2015-02-12
[Review] Request from 'schaefi' @ 'SUSE/Enceladus/azureMeta'

schaefi:
Please review the following changes:
- f762876 Added support for internal/external ip discovery
- 16bd5e5 - azuremetadata   + Make the code public   + Rename executable and package to azuremetadata

### DIFF
--- a/azuremetadata/azuremetadata.spec
+++ b/azuremetadata/azuremetadata.spec
@@ -29,6 +29,7 @@ Group:          System/Management
 Source0:        %{name}-%{version}.tar.bz2
 Requires:       perl
 Requires:       perl-JSON
+Requires:       perl-XML-LibXML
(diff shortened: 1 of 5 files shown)


schaefi:
@rjschwei, @bear454  added external and internal ip discovery, maybe you and/or James can review

rjschwei:
Looks good to me with the exception of the AzureAgent file name. While the Agent provides the file I don't think we want to tie ourselves to this by name. Lets use functional names. You used AzureDisk as an indication that this contains operations that get data from the Disk, lets use AzureNetwork to indicate the functionality this module provides is related to the network.

schaefi:
yes, makes sense will change it

rjschwei:
Looks like a re-base is required.

schaefi:
merging now before somebody else pushes changes to the old code

PR #6, opened 2015-02-17
GCE mdata

rjschwei:


gcemetadata/gcemetadata
```diff
@@ -0,0 +1,190 @@
+#!/usr/bin/env python
+#
+# Copyright (c) 2015 SUSE LLC, Robert Schweikert <rjschwei@suse.com>
+#
+# This file is part of ec2metadata.
+#
+# ec2metadata is free software; you can redistribute it and/or modify it
```
schaefi: ec2metadata typo ?

gcemetadata/lib/gcemetadata/gcemetautils.py
```diff
@@ -0,0 +1,61 @@
+#
+# Copyright (c) 2015 SUSE LLC, Robert Schweikert <rjschwei@suse.com>
+#
+# This file is part of gcemetadata.
+#
+# ec2metadata is free software: you can redistribute it and/or modify
```
schaefi, line 6: typo ?

gcemetadata/lib/gcemetadata/gcemetautils.py
```diff
@@ -0,0 +1,61 @@
...
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ec2metadata is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ec2metadata.  If not, see <http://www.gnu.org/licenses/>.
```
schaefi, line 17: typo ?

gcemetadata/lib/gcemetadata/__init__.py
```diff
@@ -0,0 +1,268 @@
...
+# (at your option) any later version.
+#
+# gcemetadata is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ec2metadata.  If not, see <http://www.gnu.org/licenses/>.
+
```
schaefi: Is there a reason why GCEMetadataException and GCEMetadata classes are implemented in one file named **init**.py ? Maybe place an empty  **init**.py to let the application know the path and put exceptions and metadata classes in its own files.

schaefi:
apart from **init.py** which I was not clear about why it contains two class implementations the code looks good and worked for me. It does more than just handling the netdev response as needed by the suma appliance but I think you will have a use case for it on another card/project

rjschwei:
Fixed the ec2 cut and paste errors, thanks for catching those.

rjschwei:
Yes, sine the framework provide a web API it is just as easy to implement query functionality that provides information for all options as it is to implement query functionality for one option. Well it's not just as easy, but it is the right thing to do ;)

rjschwei:
Since the exception is basically empy, i.e. has no special implementation putting it in it's own file is a bit overkill, IMHO, but we can certainly go that route.

Putting the implementation of the class into  **init** saves one step of indirection, but again I am OK with breaking this apart.

schaefi:
It's easier to read and you don't know if there might be more exceptions in the future. I also think refactoring code becomes easier that way. Thanks much

rjschwei:
OK, re-factor complete.

schaefi:
looks great now

PR #7, opened 2015-02-21
Default query

rjschwei:
- gcemetadata
  - Add default query behavior, if no query options are specified
    all data is displayed


schaefi:
looks good to me

PR #9, opened 2015-03-01
Clean up and logging improvement

rjschwei:


rjschwei:
No reason, fixed. String substitution is certainly not necessary here.

PR #13, opened 2015-03-25
Fetch an instance's unique ID

bear454:
... treat the name as the name; using the Incarnation tag which is unique.

### DIFF
--- a/azuremetadata/lib/AzureMetadata/AzureGeneral.pm
+++ b/azuremetadata/lib/AzureMetadata/AzureGeneral.pm
@@ -20,13 +20,71 @@ use warnings;
 use XML::LibXML;
 
 our @ISA    = qw (Exporter);
-our @EXPORT_OK = qw (get_instance_id);
+our @EXPORT_OK = qw (
+  get_instance_name
+  get_instance_id
+);
 
 sub get_instance_id {
-    my $xml = shift;
-    my @instances = $xml->getElementsByTagName('Instance');
-    my $id = $instances[0]->getAttribute('id');
-    return $id;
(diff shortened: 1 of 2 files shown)


schaefi:
The idea is nice and the number is unique I agree on that. However if we do it that way a user is now forced to run azuremetada in order to capture the admin password for suma. I think that's a step back because it should be possible to get that information from the web management console as well.

I have also mixed feelings to poke for values in the smbios table and use them as user data. it's not really obvious to a user how to dump the dmi table and that should also be well documented, especially that we also care for the machine endian when we process the information

Our current documentation explicitly states to change the password we initially set. From my point of view this one adds too much complexity

azuremetadata/lib/AzureMetadata/AzureGeneral.pm
```diff
@@ -20,13 +20,71 @@ use warnings;
...
 
 our @ISA    = qw (Exporter);
-our @EXPORT_OK = qw (get_instance_id);
+our @EXPORT_OK = qw (
+  get_instance_name
+  get_instance_id
+);
 
 sub get_instance_id {
+  my $id = fixup_guid_endianness(`dmidecode | grep UUID | cut -c8-`);
```
rjschwei, line 29: Should not execute a system call as argument to a function. This provides no opportunity to deal with any potential errors.

rjschwei:
See inline comment for suggested changes.

Overall I am OK with this. We'll just change the manager build to use the --instance-name.

What bothers me is that the name is also labeled as ID in the XML, but I guess I just have to get over that one.

PR #16, opened 2015-04-06
DO not force kill zypper

rjschwei:


rjschwei:
This is already in the maintenance stream, merging. I got ahead of myself, or behind, depending on how one looks at it.

PR #18, opened 2015-04-07
Instance name

bear454:
Prior versions referred to the instance's name via the `--instance-id` parameter, based on a poorly named attribute in waagent's XML.

### DIFF
--- a/azuremetadata/Makefile.PL
+++ b/azuremetadata/Makefile.PL
@@ -4,5 +4,5 @@ use ExtUtils::MakeMaker;
 WriteMakefile(
     EXE_FILES => ['usr/sbin/azuremetadata'],
     NAME      => 'AzureMetadata',
-    VERSION   => '3.2.1'
+    VERSION   => '4.0.0'
 );
--- a/azuremetadata/azuremetadata.spec
+++ b/azuremetadata/azuremetadata.spec
@@ -20,7 +20,7 @@ Name:           azuremetadata
 # For compatibility, remove provides 11/11/2015
 Provides:       azureMetaData
 Obsoletes:      azureMetaData <= 2.0.1
-Version:        3.2.1
+Version:        4.0.0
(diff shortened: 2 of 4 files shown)


schaefi:
looks great to me

bear454:
Tested in an Azure VM, confirmed command help, and both long and short options:

```
jfm-test-suse-sles-11-sp3-v20150330:~ # azuremetadata -n
{
   "instance_name" : "jfm-test-suse-sles-11-sp3-v20150330"
}jfm-test-suse-sles-11-sp3-v20150330:~ # azuremetadata --instance-name
{
   "instance_name" : "jfm-test-suse-sles-11-sp3-v20150330"
}jfm-test-suse-sles-11-sp3-v20150330:~ # azuremetadata 
Usage:
  azuremetadata 
    --cloudservice | -c
      print the cloud service DNS name

    --external-ip | -e
      print public external IPv4 address

    --instance-name | -n
      print the instance's unique name

    --internal-ip | -i
      print private internal IPv4 address

    --tag | -t
      print vhd disk tag stored in boot record

  options
    [ --device | -d <image file or disk device> ]
      Specify device for tag discovery. Default device
      is set to '/dev/sda'
    [ --xml | -x ]
      Wrap information in XML markup
    [ --bare | -b ]
      Just print information without markup
--
no command given at /usr/sbin/azuremetadata line 101.
```

schaefi:
as discussed on irc, there is no default action defined for azuremetadata, a better error message saying that the user has to tell what he wants is appreciated yes. But so far this is the exptected behavior

rjschwei:
Lets fix the message in the case no command is given.

"Please provide a query option"

Also do not use "die" as this gives the line number etc. Just print the message and exit with status 1. also the message should be printed to STDERR not STDOUT.

rjschwei:
The rest looks good to me.

PR #29, opened 2015-06-18
Fixed behaviour and added new instance information

schaefi:
- The --instance-name option should return the name and not the incarnation tag.
- A new option --instance-tag returns the incarnation tag
- A new option --instance-id returns the instance id as representation of the smbios UUID

### DIFF
--- a/azuremetadata/lib/AzureMetadata/AzureGeneral.pm
+++ b/azuremetadata/lib/AzureMetadata/AzureGeneral.pm
@@ -20,13 +20,78 @@ use warnings;
 use XML::LibXML;
 
 our @ISA    = qw (Exporter);
-our @EXPORT_OK = qw (get_instance_name);
+our @EXPORT_OK = qw (
+    get_instance_name
+    get_instance_incarnation_tag
+    get_instance_id
+);
 
 sub get_instance_name {
     my $xml = shift;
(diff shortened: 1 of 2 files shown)


schaefi:
> Are you confident on this?

No I'm not, do you know a more robust way to get to the instance name ?

> so maybe we just skip this for now?

which step do you mean ? I think the only mandatory thing here is that we have a robust way to get the instance name, as this is the information we actively use for the suse manager appliance
I'm open to skip everything else :)

rjschwei:
I'd say leave it. In a couple of month we will be re-writing most of this to use some other method to get the data anyway.

bear454:
> > Are you confident on this?
> 
> No I'm not, do you know a more robust way to get to the instance name ?

That was part of my motivation to use the `incarnation` tag ;-)

> > so maybe we just skip this for now?
> 
> which step do you mean ? I think the only mandatory thing here is that we have a robust way to get the instance name, as this is the information we actively use for the suse manager appliance
> I'm open to skip everything else :)

Trying to reorder/unpack the data in the guid.

schaefi:
sorry all this is just confusing and feels like waste of time

what do we want to achieve here ?

because this shouldn't be a discussion for a code review I'll take this offline for next week and close
this request. The branch I'll keep for the case that any part of this code is user relevant

PR #43, opened 2015-07-04
pint initial implementation

rjschwei:


rjschwei:
Can I get some eyes on this pull request please. 

Thanks.

schaefi:
I'm looking into this, could you review the open pull request on azurectl ? Thanks

susePublicCloudInfoClient/Makefile
```diff
@@ -7,7 +7,9 @@ files = Makefile README.md LICENSE pint setup.py
 
 nv = $(shell rpm -q --specfile --qf '%{NAME}-%{VERSION}\n' *.spec)
 verSpec = $(shell rpm -q --specfile --qf '%{VERSION}' *.spec)
-verSrc = $(shell cat lib/susepubliccloudinfoclient/VERSION)
+verSrc = $(shell cat lib/susepubliccloudinfoclient/version.py | \
+	grep __version__ | awk -F\' '{ print $$2 }' | tr -d '\n')
+
```
schaefi, line 12: you could use python to read the version

```
$(shell python -c 'from susepubliccloudinfoclient/version import __version__; print __version__')
```

afaik the coding style also suggest big letters for private global variables **version** vs. **VERSION**

But I'm fine with the above code too

susePublicCloudInfoClient/setup.py
```diff
@@ -24,7 +24,13 @@
     sys.stderr.write('Python setuptools required, please install.')
     sys.exit(1)
 
-version = open('lib/susepubliccloudinfoclient/VERSION').read().strip()
```
schaefi, line 27: you can do this with an import

```
from usepubliccloudinfoclient/version import __version__

config = {
    'version': __version__,
}
```

susePublicCloudInfoClient/lib/susepubliccloudinfoclient/infoserverrequests.py
```diff
@@ -0,0 +1,107 @@
...
+# You should have received a copy of the GNU General Public License
+# along with susePublicCloudInfoClient. If not, see
+# <http://www.gnu.org/licenses/>.
+#
+
+import requests
+
+
+def __form_url(
+        framework,
```
schaefi, line 26: maybe you could consider to make this a class:

```
class InfoServerRequests:
```

just an idea to have a namespace for this collection

susePublicCloudInfoClient/tests/getdatatest.py
```diff
@@ -0,0 +1,70 @@
...
+    expected = 'https://susepubliccloudinfo.suse.com/v1/'
+    expected += 'amazon/us-east-1/images/active.json'
+    assert_equals(expected, url)
+
+def test_form_url_images_all_xml():
+    """Form URL for all images in XML format"""
+    url = ifsrequest.__form_url('google', 'images')
+    expected = 'https://susepubliccloudinfo.suse.com/v1/google/images.xml'
+    assert_equals(expected, url)
+
```
schaefi, line 69: The tests indicates these are all public interfaces, however each test calls a method marked private
e.g __get_api_version. Is that by intention ? code considered private shouldn't need an explicit test

susePublicCloudInfoClient/Makefile
rjschwei, line 12: I didn't want to fire up the python interpreter to figure out the version, although I have to admit that awk is pretty archaic, what can I say, I am old ;)

Will fix the name of the variable, want to stick as close to pep 8 as possible.

susePublicCloudInfoClient/setup.py
rjschwei, line 27: Ooops thought I had fixed that. Thanks

susePublicCloudInfoClient/lib/susepubliccloudinfoclient/infoserverrequests.py
rjschwei, line 26: I didn't see the need for a class as there is no inherent object data, it's really just a bunch of methods tied together with a number of calls. Thus I decided to skip the overhead of having a class implementation.

susePublicCloudInfoClient/tests/getdatatest.py
rjschwei, line 69: The "public functions", i.e. function intended for call from the command line processor are only the get_image_data and get_server_data functions. everything else is intended to be used by those two functions only. 

However, testing those two functions directly ends up in network access, i.e. not testable in OBS if we care to do so. Thus I decided to test the critical for functionality implementations, and that's the generation of the URL at this point.

rjschwei:
Fixed the VERSION naming, imported the version into setup.py

PR #44, opened 2015-07-10
[ec2utilsutils] import sys and fix exc_info() call

ChrisBr:
Otherwise it'll will return in a global name 'sys' is not defined exception


ChrisBr:
Arghs, I will get rid of the Merge pull request commit ...

ChrisBr:
see #45

PR #46, opened 2015-07-10
Fix blockdev call

ChrisBr:
Like parted it's also necessary to call /usr/sbin/blockdev otherwise it fails with an command not found error.


ChrisBr:
Strangely the old code works for the SLE12 AMIs but not for SLE11SP3 AMIs...

rjschwei:
On 07/10/2015 07:06 AM, Christian Bruckmayer wrote:

> Strangely the old code works for the SLE12 AMIs but not for SLE11SP3 AMIs...

It's broken on SLES 11 because sudo on SLES 11 is broken. sudo on SLES
11 does not use, what could be considered the default path setup for
root which should include /usr/sbin.

## 

Robert Schweikert                   MAY THE SOURCE BE WITH YOU
Public Cloud Architect                         LINUX
rjschwei@suse.com
IRC: robjo

ChrisBr:
@rjschwei Ahh, thanks for the clarification!

PR #49, opened 2015-08-12
Pint filters

bear454:
Re: https://trello.com/c/BLdOFDta/363-publiccloudinfo-client


bear454:
Okay, the rest of the issues seem to center around a bigger flow issue in the code; I'll post here when I have a refactor that addresses those comments.

bear454:
@rjschwei please review again... I'm sure there are more issues, but I'm pretty happy with the progress.

susePublicCloudInfoClient/lib/susepubliccloudinfoclient/infoserverrequests.py
```diff
@@ -19,30 +19,112 @@
...
         framework,
         info_type,
-        data_format='xml',
-        region=None,
+        result_format='xml',
+        region='all',
         image_state=None,
-        server_type=None):
+        server_type=None,
+        apply_filters=None):
```
rjschwei: apply_filters i snot used and since we had agreed to not do any additional filtering on the server, past what is done via the API, i.e. formation of the URL I don't think we need to pass 'apply_filters' as an argument to the function

susePublicCloudInfoClient/lib/susepubliccloudinfoclient/infoserverrequests.py
```diff
@@ -19,30 +19,112 @@
...
+    return filtered_items
 
 
 def __form_url(
         framework,
         info_type,
-        data_format='xml',
-        region=None,
+        result_format='xml',
+        region='all',
```
rjschwei, line 107: Now the default is being set to 'all' but if the default is used then later in the code the value is set to 'None', why not just leave the default value at 'None'. It is a keyword argument and can be skipped if so desired.

I can see that the intent, i.e. "if no region is given we get information about all regions" is better communicated when the default value is "all". However, given that this is a function of the server API rather than this code I would tend to put this into a comment rather than have the additional condition later in the code.

susePublicCloudInfoClient/lib/susepubliccloudinfoclient/infoserverrequests.py
```diff
@@ -60,29 +142,106 @@ def __get_base_url():
 def __get_data(url):
     """Make the request and return the data or None in case of failure"""
     response = requests.get(url)
+    print url
```
rjschwei: I don't think we  want to print the url by default, maybe if there is a --verbose switch on the command line. Printing it by default will mess up automated processing of the result as it adds a line to the output that is not data.

susePublicCloudInfoClient/lib/susepubliccloudinfoclient/infoserverrequests.py
```diff
@@ -60,29 +142,106 @@ def __get_base_url():
...
+    if apply_filters:
+        return 'json'
+    elif result_format == 'plain':
+        return 'xml'
+    else:
+        return result_format
+ 
+
+def __warn(str, out=sys.stdout):
+    out.write("Warning: %s" % str)
```
rjschwei, line 211: We need the warning, as previously agreed/discussed. I am just wondering if we should somehow make certain there is a separation to the results, similar argument than with the "print" statement above.

susePublicCloudInfoClient/pint
```diff
@@ -56,8 +56,8 @@ import sys
 
 from docopt import docopt
 
-import infoserverrequests
-import version
+import lib.susepubliccloudinfoclient.infoserverrequests as ifsrequest
+import lib.susepubliccloudinfoclient.version as version
```
rjschwei, line 61: This is not going to work once installed. Once installed into sitelib the lib directory disappears, it is just an artifact for the code organization in development. Installed on a system all code will live only under the susepubliccloudinfoclient directory, which becomes the package in Python

rjschwei:
Almost there

susePublicCloudInfoClient/lib/susepubliccloudinfoclient/infoserverrequests.py
bear454: Oops; debugging remnants.

bear454: Removed in db047a4.

susePublicCloudInfoClient/pint
bear454, line 61: What would you suggest here that will work both for development & production use?

susePublicCloudInfoClient/lib/susepubliccloudinfoclient/infoserverrequests.py
bear454, line 107: Well, part was also reducing redundancy: both the `get_image_data` and `get_server_data` have `region='all'` in their params, and both were doing the same block of code to map it from `'all'` to `None`; so I moved it up to a common process.

susePublicCloudInfoClient/pint
rjschwei, line 61: I'd say just have what works after install. During development one can either just install from the working directory or set the environment such that the local lib is found first. Setting PATH or PYTHON_PATH work.

susePublicCloudInfoClient/lib/susepubliccloudinfoclient/infoserverrequests.py
rjschwei, line 107: Good catch

susePublicCloudInfoClient/pint
rjschwei, line 61: The import statement is the last issue before this can be merged

bear454:
@rjschwei documentation updated, specfile fixed based on a dependency problem I found while testing. Live tested on SLES 12 in EC2, it looks good there.

PR #50, opened 2015-08-13
ec2 utilities all at once

rjschwei:
Sorry, this turned into an ugly review. What started as a "just implement publishing" branch turned into "get the utilities done and ready for SLE" branch.


ec2utils/ec2deprecateimg/lib/ec2utils/ec2deprecateimg.py
```diff
@@ -73,30 +73,14 @@ def __init__(
     # ---------------------------------------------------------------------
     def _find_images_by_id(self, image_id, filter_replacement_image=None):
```
bear454, line 74: Why not just get rid of this function, and call `utils.find_images_by_id()` directly?

ec2utils/ec2publishimg/Makefile
```diff
@@ -0,0 +1,25 @@
...
+MANPATH=/usr/share/man
+PREFIX=/usr
+NAME=ec2publishimg
+dirs = lib man
+files = Makefile README.md LICENSE ec2publishimg setup.py
+
+nv = $(shell rpm -q --specfile --qf '%{NAME}-%{VERSION}\n' *.spec)
+verSpec = $(shell rpm -q --specfile --qf '%{VERSION}' *.spec)
+verSrc = $(shell cat lib/ec2utils/publish_VERSION)
+ifneq "$(verSpec)" "$(verSrc)"
```
bear454, line 11: This isn't specific to the code in this PR, just a general comment.

I _hate_ having the duplicate version numbers. Can we come up with a solution that only requires maintaining a single entry point for the version number?

ec2utils/ec2publishimg/ec2publishimg
```diff
@@ -0,0 +1,225 @@
...
+argparse.add_argument(
+    '-f', '--file',
+    default=os.path.expanduser('~') + os.sep + '.ec2utils.conf',
+    dest='configFilePath',
+    help='Path to configuration file, default ~/.ec2utils.conf (Optional)',
+    metavar='CONFIG_FILE'
+)
+# Note, one of the arguments in the group is required if --version is
+# not specified. However setting this behavior through the parser
+# also requiers one argument to be specified even if --version is specified
```
bear454: s/requiers/requires/

ec2utils/ec2publishimg/ec2publishimg
```diff
@@ -0,0 +1,225 @@
...
+help_msg = 'A regular expression to match the image name of the image to be '
+help_msg += 'published (Optional)'
+publish_image_condition_group.add_argument(
+    '--image-name-match',
+    dest='pubImgNameMatch',
+    help=help_msg,
+    metavar='REGEX'
+)
+help_msg = 'Comma separated list of regions for publishing, all integrated '
+help_msg += 'region sif not given (Optional)'
```
bear454: `regions if` not `region sif`

ec2utils/ec2publishimg/ec2publishimg
```diff
@@ -0,0 +1,225 @@
...
+    print version
+    sys.exit(0)
+
+# Explicit check required to to the group issue, see comment above
+if (
+        not args.pubImgID
+        and not args.pubImgName
+        and not args.pubImgNameFrag
+        and not args.pubImgNameMatch):
+    error_msg = 'ec2publishimg: error: one of the arguments '
```
bear454, line 148: This looks ripe for pulling out into a separate `errormsg` function, where the `'ec2publishimg: '` prefix could be added consistently and without repetition, and we could then redirect to `stderr` ?

ec2utils/ec2publishimg/ec2publishimg
```diff
@@ -0,0 +1,225 @@
...
+    error_msg += '--image-id --image-name --image-name-frag '
+    error_msg += '--image-name-match is required'
+    print error_msg
+    sys.exit(1)
+
+
+config_file = args.configFilePath
+config = None
+if not os.path.isfile(config_file):
+    print 'Configuration file "%s" not found.' % config_file
```
bear454, line 158: See - this one doesn't have the `'ec2publishimg: '` prefix ;-)

ec2utils/ec2publishimg/ec2publishimg
```diff
@@ -0,0 +1,225 @@
...
+    sys.exit(1)
+try:
+    config = utils.get_config(config_file)
+except Exception as e:
+    print >> sys.stderr, e.message
```
bear454, line 163: ... and here we actually do go out to `stderr`.

ec2utils/ec2publishimg/lib/ec2utils/ec2publishimg.py
```diff
@@ -0,0 +1,130 @@
...
+                    self.image_name_match)
+            except:
+                msg = 'Unable to complie regular expression "%s"'
+                msg = msg % self.image_name_match
+                raise EC2PublishImgException(msg)
+
+    # --------------------------------------------------------------------
+    def _print_image_info(self, image):
+        """Print a message about the image that would be modified"""
+        if self.visibility == 'all' or self.visibility == 'none':
```
bear454, line 85: Since you parse the `visibility` on `init`, would it hurt anything to just send the extra param along and avoid the redundant `if` block?

```
def _print_image_info(self, image):
    print self.publish_msg % (image.id, image.name, self.visibility)
```

bear454:
LOL @ `ec2utilsutils`.  Maybe we can add a utility subclass, `ec2utilsutilsutils` :wink:

ec2utils/ec2deprecateimg/lib/ec2utils/ec2deprecateimg.py
rjschwei, line 74: Because for deprecation we have the special condition that one can use the virtualization type as an additional filter. Somehow we need to trim the list of potential images to be deprecated based on this condition.

ec2utils/ec2publishimg/ec2publishimg
rjschwei, line 148: The prefix is there to stay in line with the error style argparse would produce for a missing argument. However, as noted in the comment the group cannot be handled by argparse, thus the goal was to make the error appear in the same style even though it is generated by this code rather than argparse.

ec2utils/ec2publishimg/ec2publishimg
rjschwei, line 158: This is alos not a command line usage error, two different categories.

ec2utils/ec2publishimg/ec2publishimg
rjschwei, line 163: Yes, the previous error should go to stderr

ec2utils/ec2publishimg/lib/ec2utils/ec2publishimg.py
rjschwei, line 85: Sorry I don't get where there is anything redundant. The message for publishing, i.e. visibility = all and for privatizing, i.e. visbility = none has 2 string substitutions. The sharing with specific accounts has 3 string substitutions. The tuple passed for unpacking has to be exact or it will trigger a runtime error.

bear454, line 85: OK, I assumed the additional param would simply be ignored in string substition.

PR #52, opened 2015-09-16
- Add subdirectory for Nagios/Icinga plugins (monitor)

rjschwei:
- Implement a process monitor we can use to monitor smt- processes for
  "hanging" issues. The monitor allows to set a certain runtime for a given
  process after which the process is considered hung. The process can
  optionally be killed if it's runtime is expired


bear454:
Looks good to me. Thanks for writing that!

PR #59, opened 2015-09-30
Handle failover in the client

rjschwei:


regionServiceClient/cloud-regionsrv-client.spec
```diff
@@ -132,7 +132,15 @@ rm -rf $RPM_BUILD_ROOT
 %files
 %defattr(-,root,root,-)
 %doc README LICENSE
+%dir %{_usr}/lib/zypp
+%dir %{_usr}/lib/zypp//plugins
```
bear454: Should there be two slashes here? `zypp//plugins`

regionServiceClient/lib/cloudregister/registerutils.py
```diff
@@ -0,0 +1,419 @@
...
+            continue
+        if smt_announce_found and domain_name in entry:
+            smt_announce_found = False
+            continue
+        new_hosts_content.append(entry)
+
+    hosts_file = open('/etc/hosts', 'w')
+    for entry in new_hosts_content:
+        hosts_file.write(entry)
+    hosts_file.close()
```
bear454: This is purely style, so I'm not going to get upset if you leave it, but I prefer

```
with open(HOSTSFILE, 'w') as hosts_file:
    for entry in new_hosts_content:
        hosts_file.write(entry)
```

(1) using the block for the open file, so it self-closes, and (2) abstacting the magic string out to a constant.

regionServiceClient/lib/cloudregister/registerutils.py
```diff
@@ -0,0 +1,419 @@
...
+        cmd = subprocess.Popen(
+            cmd,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE
+        )
+        cmd.communicate()
+    except:
+        return 0
+
+    return 1
```
bear454, line 78: Why `0` and `1` for returns? What do they represent?

regionServiceClient/lib/cloudregister/registerutils.py
```diff
@@ -0,0 +1,419 @@
...
+    return None
+
+
+# ----------------------------------------------------------------------------
+def get_available_smt_servers():
+    """Return a list of available SMT servers"""
+    availabe_smt_servers = []
+    if not os.path.exists(REGISTRATION_DATA_DIR):
+        return availabe_smt_servers
+    smt_data_files = glob.glob(REGISTRATION_DATA_DIR + 'availableSMTInfo*')
```
bear454, line 100: How come `REGISTRATION_DATA_DIR` is a constant, but `'availableSMTInfo*'` is inlined?

Sorry, I know, style nitpick.

regionServiceClient/lib/cloudregister/registerutils.py
```diff
@@ -0,0 +1,419 @@
...
+
+    smt_file = open(smt_store_file_path, 'r')
+    u = pickle.Unpickler(smt_file)
+    try:
+        smt = u.load()
+    except:
+        smt_file.close()
+        return None
+
+    smt_file.close()
```
bear454: What do you think about ...

```
smt = None
with open(smt_store_file_path, 'r') as smt_file:
    u = pickle.Unpickler(smt_file)
    try:
        smt = u.load()
return smt
```

regionServiceClient/lib/cloudregister/registerutils.py
```diff
@@ -0,0 +1,419 @@
...
+    if not import_result:
+        logging.error('SMT certificate import failed')
+        return None
+
+    return 1
+
+
+# ----------------------------------------------------------------------------
+def is_registered(smt):
+    """Firgure out if any of the servers is known and this instan"""
```
bear454: "Figure" typo

bear454: Don't understand "and this instan" ?

regionServiceClient/lib/cloudregister/registerutils.py
```diff
@@ -0,0 +1,419 @@
...
+    current_smt_ip = current_smt.get_ip()
+    for entry in known_hosts:
+        if current_smt_ip in entry:
+            new_hosts += '%s\t%s\t%s\n' % (new_smt.get_ip(),
+                                           new_smt.get_FQDN(),
+                                           new_smt.get_name())
+            continue
+        new_hosts += entry
+
+    hosts = open('/etc/hosts', 'w')
```
bear454: This is a different interpretation of the same write block as in `clean_hosts_file`. How about abstracting it out to its own function?

regionServiceClient/lib/cloudregister/registerutils.py
```diff
@@ -0,0 +1,419 @@
...
+    except IOError:
+        print 'Could not open log file ', logFile, ' for writing.'
+        sys.exit(1)
+
+
+# ----------------------------------------------------------------------------
+def store_smt_data(smt_data_file_path, smt):
+    """Store the given SMT server information to the given file"""
+    smt_data = open(smt_data_file_path, 'w')
+    os.fchmod(smt_data.fileno(), stat.S_IREAD | stat.S_IWRITE)
```
bear454, line 285: Should the code change the permissions back after the write is complete?

regionServiceClient/lib/cloudregister/registerutils.py
rjschwei, line 78: Failure and success

regionServiceClient/lib/cloudregister/registerutils.py
rjschwei, line 100: The latter is only used once

regionServiceClient/lib/cloudregister/smt.py
```diff
@@ -0,0 +1,151 @@
...
+
+        return True
+
+
+    # --------------------------------------------------------------------
+    def __request_cert(self):
+        """Request the cert from the SMT server and return the request"""
+        cert_rq = None
+        attempts = 0
+        retries = 3
```
bear454, line 138: why not make these default attributes on the method call?

```
def __request_cert(self, cert_rq=None, attempts=0, retries=3):
```

regionServiceClient/lib/cloudregister/registerutils.py
rjschwei: Well, not quite like that but I see where you are going and it is reasonable to shorten the implementation.

regionServiceClient/lib/cloudregister/registerutils.py
```diff
@@ -0,0 +1,419 @@
...
+        if smt.get_ip() == configured_smt.get_ip():
+            continue
+        if smt.is_equivalent(configured_smt) and smt.is_responsive():
+            return smt
+
+    return None
+
+
+# ----------------------------------------------------------------------------
+def get_available_smt_servers():
```
bear454, line 95: Looking now at how this is consumed by `regionServiceClient/usr/lib/zypp/plugins/services/cloud_update`, wouldn't it make more sense to query the regionserver again, instead of relying on the cached document, in case any of the data has changed?

( I know it hasn't changed much up to this point... maybe this is just YAGNI).

bear454:
I only had a couple of questions about the process, most comments are just style & a few typos in comments.

Overall - this is awesome. Thanks @rjschwei

bear454:
One other nitpick that I didn't want to put inline, was the return values. It seems like most of the returns are intended to be boolean, but I saw `return 1`/`return None`, `return 1`/`return 0`, `return True`/`return False`. I'd _like_ to see that more consistent.

regionServiceClient/lib/cloudregister/registerutils.py
rjschwei: Then I need to do a conversion of data type, in clean_hosts_file the content is treated in list context and here as a string. The conversion takes as many lines as it takes to keep tw implementation of the writing, vs. write_hosts_file(content) thus from my perspective it is equivalent to the current implementation.

regionServiceClient/lib/cloudregister/smt.py
rjschwei, line 138: That would imply that somewhere along the line they can be changed via arguments, which they cannot today, and I am not certain I want to go in that direction. This is only there to accommodate network flakiness of the framework and I really do not want to encourage having a work for a insufficiently responsive framework implementation.

regionServiceClient/lib/cloudregister/registerutils.py
rjschwei, line 95: I did think about querying the region server again. Doing so would have the advantage that we could do a complete rip and replace of the SMT servers with new certs etc. However, I couldn't convince myself that the additional complication was worth the uncertain amount of "future proofing" we would get. Thus I opted to stay with the simpler solution of staying with the data we had already received.

rjschwei:
Well, the return values hopefully go along with the "questions" being asked or the actions that are supposed to be performed.

a question such as is_equivalent() should return True/False 
a get_me_an_obj() method should return the object or None
a do_this_method() should return success (1) or failure (0)

IMHO treating them all the same is confusing.

Issue #60, opened 2015-09-30
pint: name filtering errors out on XML output on SLES 11


bear454:
```
# pint amazon images --active --xml --region=cn-north-1 --filter="name~sp4"
Traceback (most recent call last):
  File "/usr/bin/pint", line 102, in <module>
    command_args['--filter'])
  File "/usr/lib64/python2.6/site-packages/susepubliccloudinfoclient/infoserverrequests.py", line 240, in get_image_data
    return __process(url, info_type, command_arg_filter, result_format)
  File "/usr/lib64/python2.6/site-packages/susepubliccloudinfoclient/infoserverrequests.py", line 219, in __process
    return __reformat(filtered_items, info_type, result_format)
  File "/usr/lib64/python2.6/site-packages/susepubliccloudinfoclient/infoserverrequests.py", line 195, in __reformat
    return ET.tostring(root, 'UTF-8', 'xml')
TypeError: tostring() takes at most 2 arguments (3 given)
```

The issue is not present when using JSON output:

```
# pint amazon images --active --json --region=cn-north-1 --filter="name~sp4"
{"images": [{"deletedon": "", "name": "suse-sles-11-sp4-v20150714-pv-ssd-x86_64", "region": "cn-north-1", "replacementname": "", "deprecatedon": "", "state": "active", "replacementid": "", "publishedon": "20150714", "id": "ami-8066fbb9"}, {"deletedon": "", "name": "suse-sles-11-sp4-v20150714-hvm-ssd-x86_64", "region": "cn-north-1", "replacementname": "", "deprecatedon": "", "state": "active", "replacementid": "", "publishedon": "20150714", "id": "ami-3461fc0d"}, {"deletedon": "", "name": "suse-sles-11-sp4-byos-v20150714-pv-ssd-x86_64", "region": "cn-north-1", "replacementname": "", "deprecatedon": "", "state": "active", "replacementid": "", "publishedon": "20150714", "id": "ami-fe66fbc7"}, {"deletedon": "", "name": "suse-sles-11-sp4-byos-v20150714-hvm-ssd-x86_64", "region": "cn-north-1", "replacementname": "", "deprecatedon": "", "state": "active", "replacementid": "", "publishedon": "20150714", "id": "ami-3661fc0f"}]}
```

PR #61, opened 2015-10-01
Alter XML output to be compatible with both Python 2.6 & 2.7

bear454:
Python 2.7 introduced a third param to
xml.etree.ElementTree.tostring(element, encoding="us-ascii", method="xml")
, but it defaults to the desired value: xml.

Python 2.6 (Included in SLE 11) did not have this attribute, so the method
failed there.

Relying on the default value for Python 2.7 (SLE 12, openSUSE,
Tumbleweed) satisfies both options.


bear454:
Resolves https://github.com/SUSE/Enceladus/issues/60

PR #63, opened 2015-10-04
AWS info gen

rjschwei:
Amazon requested to have the data relevant to their operations presented in CSV format. This allows us to easily generate the data by running the new command and chaining some grep statements together. The new binary does not need to be released, thus it is in a separate package.


bear454:
Looks good to me.

bear454:
I take that back; `make tar` fails due to version detection on the now-more-complex specfile.

bear454:
```
rpm -q --specfile --qf '%{VERSION}' *.spec
=> 0.2.00.2.0
```

Just off the cuff (I'm sure you'll find something cleaner)

```
rpm -q --specfile --qf '%{VERSION}\n' *.spec | head -n1 | tr -d '\n'
=> 0.2.0
```

PR #72, opened 2016-02-04
- supportutilsPluginSUSEPublicCloud

rjschwei:
- Implement supportutil plugin for the public cloud

### DIFF
--- a/supportutilsPluginSUSEPublicCloud/supportconfig/plugins/suse_public_cloud
+++ b/supportutilsPluginSUSEPublicCloud/supportconfig/plugins/suse_public_cloud
@@ -1,0 +1,180 @@
+#!/bin/bash
+
+# Supportconfig Plugin for SUSE Public Cloud Images
+# 
+# Copyright (C) 2016 SUSE LLC, Robert Schweikert <rjschwei@suse.com>
+# 
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+# 
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+# 
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor,
+# Boston, MA  02110-1301, USA.
+
+# Gathers important troubleshooting information about
+# SUSE Public Cloud Instances
+
+SVER='1.0'
+RCFILE="/usr/lib/supportconfig/resources/scplugin.rc"
+LOG_LINES=0  # 0 means include the entire file
+
+# Collect all the public cloud relevant data in a directory
+DATA_LOCATION="$LOG/public_cloud"
+mkdir -p $DATA_LOCATION
+
+if [ -s $RCFILE ]; then
+    if ! source $RCFILE; then
+        echo "ERROR: Initializing resource file: $RCFILE" >&2
+        exit 1
+    fi
+fi
+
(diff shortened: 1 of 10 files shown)


supportutilsPluginSUSEPublicCloud/supportconfig/plugins/suse_public_cloud
```diff
@@ -0,0 +1,180 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor,
+# Boston, MA  02110-1301, USA.
+
+# Gathers important troubleshooting information about
+# SUSE Public Cloud Instances
+
+SVER='1.0'
+RCFILE="/usr/lib/supportconfig/resources/scplugin.rc"
```
schaefi, line 26: It's good to stay consistent with the quoting in bash, the policy here are " quotes, I would avoid ' quotes

supportutilsPluginSUSEPublicCloud/supportconfig/plugins/suse_public_cloud
```diff
@@ -0,0 +1,180 @@
...
+LOG_LINES=0  # 0 means include the entire file
+
+# Collect all the public cloud relevant data in a directory
+DATA_LOCATION="$LOG/public_cloud"
```
schaefi, line 30: when concatenate from variable content it most safe to use the ${LOG} notation to avoid unexpected results

supportutilsPluginSUSEPublicCloud/supportconfig/plugins/suse_public_cloud
```diff
@@ -0,0 +1,180 @@
...
+mkdir -p $DATA_LOCATION
+
+if [ -s $RCFILE ]; then
```
schaefi, line 33: always quote in tests: if [ -s "$RCFILE" ]

supportutilsPluginSUSEPublicCloud/supportconfig/plugins/suse_public_cloud
```diff
@@ -0,0 +1,180 @@
...
+    if ! source $RCFILE; then
+        echo "ERROR: Initializing resource file: $RCFILE" >&2
+        exit 1
+    fi
+fi
+
+_collect_all_file_data() {
```
schaefi, line 40: bash style would normally use the function keyword, I have no strong opinion though

```
function _collect_all_file_data {
}
```

supportutilsPluginSUSEPublicCloud/supportconfig/plugins/suse_public_cloud
```diff
@@ -0,0 +1,180 @@
...
+    dir=$1
+    dest_file=$2
+    for f in $dir/*; do
```
schaefi, line 43: You want globbing, I suggest to explicitly activate that

set +o noglob

There is no guarantee the default bash setup provides it, however our shell setup always has it switched on

supportutilsPluginSUSEPublicCloud/supportconfig/plugins/suse_public_cloud
```diff
@@ -0,0 +1,180 @@
...
+    _collect_all_file_data '/etc/zypp/credentials.d' $INFO_FILE
+}
+
+framework() {
+    local INFO_FILE="$DATA_LOCATION/framework.txt"
+    echo "#==[ Determine Framework ]================================#" > $INFO_FILE
+    echo "# dmidecode | grep -q amazon|Google|Microsoft" >> $INFO_FILE
+
+    dmidecode | grep -q amazon
+    if (( $? == 0 )); then
```
schaefi, line 63: You can also write this as

```
if dmidecode | grep -q amazon; then
   ...
fi
```

schaefi:
I just saw this is already merged, I guess a review might make not much sense any more ?

rjschwei:
Review always makes sense, it's just code and can always be changed :) thanks allot for the comments and the review.

Issue #75, opened 2016-02-23
pint displays traceback


rjschwei:
If the server cannot be reached pint displays a traceback

Traceback (most recent call last):
  File "/usr/bin/pint", line 102, in <module>
    command_args['--filter'])
  File "/usr/lib/python2.7/site-packages/susepubliccloudinfoclient/infoserverrequests.py", line 235, in get_image_data
    return __process(url, info_type, command_arg_filter, result_format)
  File "/usr/lib/python2.7/site-packages/susepubliccloudinfoclient/infoserverrequests.py", line 211, in __process
    server_response_data = __get_data(url)
  File "/usr/lib/python2.7/site-packages/susepubliccloudinfoclient/infoserverrequests.py", line 139, in __get_data
    response = requests.get(url)
  File "/usr/lib/python2.7/site-packages/requests/api.py", line 69, in get
    return request('get', url, params=params, *_kwargs)
  File "/usr/lib/python2.7/site-packages/requests/api.py", line 50, in request
    response = session.request(method=method, url=url, *_kwargs)
  File "/usr/lib/python2.7/site-packages/requests/sessions.py", line 468, in request
    resp = self.send(prep, *_send_kwargs)
  File "/usr/lib/python2.7/site-packages/requests/sessions.py", line 576, in send
    r = adapter.send(request, *_kwargs)
  File "/usr/lib/python2.7/site-packages/requests/adapters.py", line 420, in send
    raise ConnectionError(e, request=request)
requests.exceptions.ConnectionError: HTTPSConnectionPool(host='susepubliccloudinfo.suse.com', port=443): Max retries exceeded with url: /v1/amazon/us-east-1/images/active.json (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x7fda855d2b50>: Failed to establish a new connection: [Errno 111] Connection refused',))

rjschwei:
There is another one.

pint amazon images --active --region us-east --filter 'name~12-sp1,name~hvm'
Traceback (most recent call last):
  File "/usr/bin/pint", line 102, in <module>
    command_args['--filter'])
  File "/usr/lib/python2.7/site-packages/susepubliccloudinfoclient/infoserverrequests.py", line 235, in get_image_data
    return __process(url, info_type, command_arg_filter, result_format)
  File "/usr/lib/python2.7/site-packages/susepubliccloudinfoclient/infoserverrequests.py", line 211, in __process
[...]

PR #76, opened 2016-02-24
Show a human-readable error message on request issues instead of a traceback

bear454:
re: #75


susePublicCloudInfoClient/lib/susepubliccloudinfoclient/infoserverrequests.py
```diff
@@ -205,6 +220,11 @@ def __warn(str, out=sys.stdout):
     out.write("Warning: %s" % str)
 
 
+def __error(str, out=sys.stdout):
```
rjschwei: Error should be written to stderr, since the method is called __error that should be the default

bear454: Fixed in c340e36 .

PR #77, opened 2016-02-29
Pint regex filter

jgleissner:
Support for regex and inverse substring matches.


bear454:
:+1: It's great to see some useful extension of the the filters!

bear454:
One thing @jgleissner: you'll need to bump the minor version as well, according to semantic versioning.

rjschwei:
Agreed lets bump the version, then file a bug against SLE and get a new package released. We shouldn't hold onto this change in the code only. Also rebuild the package in OBS.

jgleissner:
Sorry, wasn't aware of that. Version bumped. I'll merge and update the packages.

PR #79, opened 2016-03-08
fix traceback on HTTP error (issue #75)

jgleissner:


susePublicCloudInfoClient/lib/susepubliccloudinfoclient/infoserverrequests.py
```diff
@@ -163,6 +163,7 @@ def __get_data(url):
     """Make the request and return the data or None in case of failure"""
     try:
         response = requests.get(url)
+        response.raise_for_status()
```
rjschwei: That works but then...

PR #80, opened 2016-03-15
- serviceAccessConfig

rjschwei:
- Implement config generator for services.
  With this initial implementation we support Apache and HAProxy access
  rule generation plugins.


serviceAccessConfig/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Contributing
+
+Any submissions need to be accompanied by a test.
+
+Code needs to be pep8 clean
+
+### Implementing plugins
+
+PLugin names are linked by name to section names in the configuration file.
```
bear454: "Plugin", not "PLugin"

serviceAccessConfig/README.md
```diff
@@ -0,0 +1,159 @@
...
+For example the plugin for Apache is named "apache.py" and the expected
+section in the configuration file is named "[apache]". The class name of the
+plugin is also linked to the section name in that the class name always starts
+with "ServiceAccessGenerator" followed by the section name with a capitalized
+first letter, i.e. for Apache the class names is "ServiceAccessGeneratorApache".
+The generator class must inherit from "ServiceAccessGenerator" and must
+implement the "_update_service_config()" method
```
bear454: Add a period at the end.

serviceAccessConfig/etc/init.d/serviceAccessConfig
```diff
@@ -0,0 +1,84 @@
...
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+### BEGIN INIT INFO
+# Provides:          cspInfraServerAccessConfig
+# Required-Start:    $apache2
```
bear454: Does this make sense if Apache isn't in use? For example, if configuration only manages HAProxy?

serviceAccessConfig/lib/serviceAccessConfig/apache.py
```diff
@@ -0,0 +1,147 @@
...
+import ConfigParser
+import glob
+import logging
+import os
+import re
+
+from accessrulegenerator import ServiceAccessGenerator
+from generatorexceptions import *
+
+apache_directory_directive = re.compile(r'\s*<Directory .*>')
```
bear454, line 31: Why is this in the global space?

serviceAccessConfig/tests/unit/test_apache.py
```diff
@@ -0,0 +1,231 @@
...
+    )
+    gen_result = open(gen_result_file).read()
+
+    if ref_result == gen_result:
+        # success
+        utils.remove_test_tmpdir()
+    else:
+        msg = 'Test failed, not removing test directory '
+        msg += '"%s" to aid debugging ' % utils.get_test_tmpdir()
+        assert 1 == 0, msg
```
bear454: `assert False, msg`

serviceAccessConfig/tests/unit/test_base.py
```diff
@@ -0,0 +1,194 @@
...
+    """Test that the proper systemd command is issued to start a service"""
+
+    gen = ServiceAccessGenerator('%s/ip_data.cfg' % utils.get_data_path())
+    gen.service_name = 'foo'
+    with pytest.raises(ServiceAccessGeneratorServiceRestartError) as excinfo:
+        gen._restart_service()
+
+    assert mock_logging.error.called
+    mock_logging.error.assert_called_with('foo restart failed')
+    mock_os.system.assert_called_with('systemctl restart foo.service')
```
bear454, line 99: It would be better to mock out the point of system interaction, instead of trying to run the service, failing, and looking for the error. I'm tempted to add a foo.service to my system just to force a test failure :P

serviceAccessConfig/tests/unit/test_genfactory.py
```diff
@@ -0,0 +1,100 @@
...
+    with pytest.raises(ServiceAccessGeneratorConfigError) as excinfo:
+        generators = get_access_rule_generators(config)
+
+    assert mock_logging.error.called
+
+
+# ======================================================================
+@patch('serviceAccessConfig.generatorfactory.logging')
+def test_factory_missing_option(mock_logging):
+    """Test that an incomplte configuration file, with the missing 'ipdata'
```
bear454: _incomplete_

serviceAccessConfig/usr/sbin/serviceAccessConfig
```diff
@@ -0,0 +1,108 @@
...
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+"""Command serviceAccessConfig
+
+Usage:
+    serviceAccessConfig  -h | --help
+    serviceAccessConfig [ --config --ipdata --log ]
```
bear454: I don't think that does what you think it does [1]... you're going to end up with boolean options.

I think you want something like

```
serviceAccessConfig [--config=<config-file-name>] [--ipdata=<ip-data-file-name>] [--log=<log-file-name>]
```

[1] http://try.docopt.org/?doc=Command+serviceAccessConfig%0D%0AUsage%3A%0D%0A++++serviceAccessConfig++-h+|+--help%0D%0A++++serviceAccessConfig+[+--config+--ipdata+--log+]%0D%0A++++serviceAccessConfig+--version%0D%0AOptions%3A%0D%0A++++--config%0D%0A++++++++The+configuration+file+for+serviceAccessConfig%0D%0A++++-h+--help%0D%0A++++++++Show+help%0D%0A++++--ipdata%0D%0A++++++++The+source+data+file+to+monitor+from+which+the+CIDR+blocks+will%0D%0A++++++++be+extracted+to+generate+the+access+data+for+the+configured+service%28s%29%0D%0A++++--log%0D%0A++++++++The+log+file+to+be+written&argv=--config%3D%22test.cfg%22

serviceAccessConfig/lib/serviceAccessConfig/accessrulegenerator.py
```diff
@@ -0,0 +1,186 @@
...
+
+import ConfigParser
+import hashlib
+import logging
+import os
+import requests
+import sys
+import threading
+
+from generatorexceptions import *
```
bear454, line 30: Don't let @schaefi see that; [he'll get upset.](https://github.com/SUSE/azurectl/pull/105/files/b2ade0de81ef386cc2e42dc67a842d345c059146#diff-b13dabd37635a09d6edef0e6c0ab4406R17) ;-)

serviceAccessConfig/lib/serviceAccessConfig/accessrulegenerator.py
```diff
@@ -0,0 +1,186 @@
...
+    def update_config(self):
+        """Update the Apache configuration file"""
+        ctx = open(self.ip_source_config_file_name, 'r').read()
+        sha1 = hashlib.sha1()
+        sha1.update(ctx)
+        if (
+                not self.ip_source_config_file_sha1 or
+                self.ip_source_config_file_sha1 != sha1.digest()
+        ):
+            self.ip_source_config_file_sha1 = sha1.digest()
```
bear454, line 104: Does this get persisted anywhere?

serviceAccessConfig/lib/serviceAccessConfig/apache.py
```diff
@@ -0,0 +1,147 @@
...
+            upgradeScript = glob.glob('/usr/share/apache2/*-upgrade')
+            for scr in upgradeScript:
+                toVer = scr.split('-')[2]
+                if int(toVer) >= 24:
+                    directive = 'Require ip'
+                    break
+            return directive
+
+        apacheInfo = os.popen('%s -V' % apachectl).readlines()
+        for ln in apacheInfo:
```
bear454, line 65: Consider doing a regex across `apacheInfo` instead?

bear454, line 65: Like `/version: Apache\/(\d+)\.(\d+)\.(\d+) /` ?

serviceAccessConfig/lib/serviceAccessConfig/apache.py
```diff
@@ -0,0 +1,147 @@
...
+    # ======================================================================
+    def _update_service_config(self, cidr_blocks):
+        """Update the Apache configuragtion file"""
+
+        order = '        Order allow,deny\n'
+        cidr_blocks = cidr_blocks.split(',')
+        allow = ''
+        space = ' '
+        ip_directive = self._get_apache_ip_directive()
+        while len(cidr_blocks) > 300:
```
bear454, line 85: Are you testing with any long strings of cidr blocks?

serviceAccessConfig/etc/init.d/serviceAccessConfig
rjschwei: To the best of my knowledge insserv or chkconfig do not error out is the service is not on the system, i.e. if it is an haproxy system and apache is not installed this is OK. I did add haproxy to required start as support for both is implemented.

Also whild looking at this I noticed the cut and paste error in Provides: which I fixed as well.

serviceAccessConfig/lib/serviceAccessConfig/apache.py
rjschwei, line 31: Because I want the expression to be compiled only once and I did not want to make it a member of the class.

serviceAccessConfig/lib/serviceAccessConfig/accessrulegenerator.py
rjschwei, line 104: Nope, it stays in memory only, i.e. on restart of this service we will always force a restart of the dependent service.

serviceAccessConfig/lib/serviceAccessConfig/apache.py
rjschwei, line 65: I see no advantage for a regex at this point as I still have to pull out the the matches to form the int for version comparison.

serviceAccessConfig/lib/serviceAccessConfig/apache.py
rjschwei, line 85: Nope, only in production ;)

serviceAccessConfig/lib/serviceAccessConfig/apache.py
bear454, line 65:
```
version = float(re.search(r'Apache\/(\d+\.\d+)\.', apacheInfo).group(1))
if version >= 2.4
   ....
```

:P

PR #81, opened 2016-04-07
Convert ec2utils to be boto3 based

rjschwei:


ec2utils/ec2utilsbase/lib/ec2utils/ec2utilsutils.py
```diff
@@ -175,21 +174,26 @@ def get_from_config(account, config, region, entry, cmd_line_arg):
...
     if command_args.regions:
         regions = command_args.regions.split(',')
     else:
         regions = []
-        regs = boto.ec2.regions()
+        regs = boto3.client(
+                    aws_access_key_id=access_key,
+                    aws_secret_access_key=secret_key,
+                    region_name='us-east-1',
+                    service_name='ec2').describe_regions()['Regions']
```
bear454, line 189:
```
return boto3.session.Session(
    aws_access_key_id=access_key,
    aws_secret_access_key=secret_key
).get_available_regions('ec2')
```

This doesn't include disconnected regions by default, so no need to filter the out after ;-)

http://boto3.readthedocs.org/en/latest/reference/core/session.html?highlight=region#boto3.session.Session.get_available_regions

rjschwei, line 189: Great find, I created an issue: https://github.com/SUSE/Enceladus/issues/82

At this point I do not want to handle a boto3 version upgrade along with the conversion.

ec2utils/ec2utilsbase/lib/ec2utils/ec2utilsutils.py
```diff
@@ -76,10 +75,10 @@ def find_images_by_name(images, image_name):
     """Return a list of images that match the given name."""
     matching_images = []
     for image in images:
-        if not image.name:
+        if not image['Name']:
```
bear454, line 78: How did you end up with dicts ? You should have a collection of [`Image`](https://boto3.readthedocs.org/en/latest/reference/services/ec2.html#image) from [`EC2.ServiceResource#images`](https://boto3.readthedocs.org/en/latest/reference/services/ec2.html#EC2.ServiceResource.images)

ec2utils/ec2deprecateimg/lib/ec2utils/ec2deprecateimg.py
```diff
@@ -120,27 +119,29 @@ def _get_all_type_match_images(self, filter_replacement_image=None):
         """Get all images that match thespecified virtualization type.
            All images owned by the account if not type is specified."""
         images = []
-        my_images = self.ec2.get_all_images(owners='self')
+        my_images = self.ec2.describe_images(Owners=['self'])['Images']
```
bear454: Ooh. Did you want to use Client#describe_images, and work with the dicts, or did you want to stick with objects & attributes, using the resource?

ec2utils/ec2utilsbase/lib/ec2utils/ec2utils.py
```diff
@@ -63,6 +63,14 @@ def _disconnect_from_ec2(self):
             del self.ec2
             self.ec2 = None
 
+
+    # ---------------------------------------------------------------------
+    def _get_owned_images(self):
+        """Return the list of images owned by the account used for
+           uploading"""
+        return self.ec2.describe_images(Owners=['self'])['Images']
```
bear454: Again...

```
boto3.resource('ec2').images.filter(Owners=['self'])
```

https://boto3.readthedocs.org/en/latest/reference/services/ec2.html#EC2.ServiceResource.images

rjschwei:
I decided to use the Client approach as it is more straight forward rather than the many different layers sessions and resources represent. With that decision made I also decided that I wanted to stay consistent, i.e. use the client model approach rather than using resources for some things and clients for others. While using the client model feels somewhat un-pythonic as one has to deal with primitives it is reasonably straight forward and consistent.

PR #84, opened 2016-04-08
- ec2utils  -- ec2publishimg

rjschwei:
- Support image copy permission. An image can only be copied into
  the account with which the image is shared if the underlying
  snapshot is also shared with the account. The --allow-copy command
  line option will open the permission of the underlying snapshot


bear454:
:+1:

PR #86, opened 2016-04-12
- regionServiceClient

rjschwei:
- If the initial base product registration fails in SLE 12 or
  if initial registration fails in SLE 11 try the other update servers
  that were supplied by the region server


bear454:
Looks good, merging.

PR #88, opened 2016-06-07
- regionServiceClient

rjschwei:
- Add new configuration option use_metadata_server to use a
  metadata server for obtaining the SMT information rather than a
  region server. This is intended for use with SOC


regionServiceClient/etc/regionserverclnt.cfg
```diff
@@ -2,6 +2,7 @@
 api = regionInfo
 certLocation = /var/lib/regionService/certs
 regionsrv = COMMA_SEP_LIST_OF_CLOUD_SPECIFIC_REGION_SERVER
+use_metadata_server = OPTIONAL_URL_FOR_METADATA_SERVER_SMT_INFO
```
bear454, line 5: `use_metadata_server` sounds like it takes a boolean argument. Consider just `metadata_server` instead?

regionServiceClient/usr/sbin/registercloudguest
```diff
@@ -193,45 +187,87 @@ if proxy:
...
+                'Metadata server returned %s' % response.status_code
+            )
+    except Exception as e:
+        logging.error('=' * 20)
+        logging.error(e.message)
+        logging.error('Unable to obtain SMT server information, exiting')
+        sys.exit(1)
+    smt_info = json.loads(response.text)
+    expected_entries = ('fingerprint', 'SMTserverIP', 'SMTserverName')
+    smt_info_xml = '<regionSMTdata><smtInfo '
```
bear454, line 215: Why not use something like [SimpleXMLWriter](http://effbot.org/zone/xml-writer.htm) ? Manually constructing XML as a string is pretty error-prone.

rjschwei:
Option name changed as suggested. Decided to continue to construct the XML string.

PR #89, opened 2016-06-13
ec2uploadimg: set SRIOV for hvm images

jgleissner:
Automatically set SRIOV to 'simple' for hvm virtualization (fixes
bsc#981224).


ec2utils/ec2uploadimg/ec2uploadimg
```diff
@@ -296,6 +296,9 @@ virtualization_type = args.virtType
 if virtualization_type == 'para':
     virtualization_type = 'paravirtual'
 
+if virtualization_type == 'hvm' and not sriov_type:
+    sriov_type = 'simple'
```
rjschwei, line 300: Nope, this is not the solution to this problem. SRIOV requires a kernel module. While we have this module there is no guarantee that someone building their own image includes the module. Thus an image that gets this set automatically but does not actually have the module will have no network as the framework uses SRIOV enable HW instead of non SRIOV enabled HW.

The solution is a change in the way the image registration command is issued.

jgleissner:
Sorry, I didn't know that there was a kernel module required. However, the AWS cli documentation (http://docs.aws.amazon.com/cli/latest/reference/ec2/register-image.html) reads:

```
--sriov-net-support (string)

Set to simple to enable enhanced networking for the AMI and any instances that you launch from the AMI.
There is no way to disable enhanced networking at this time.
This option is supported only for HVM AMIs. Specifying this option with a PV AMI can make instances launched from the AMI unreachable.
```

As I understand it, SRIOV is required for HVM instances, which would mean that images without SRIOV kernel support don't work as HVM images.

rjschwei:
OK, it could be that the underlying implementation has changed in EC2, although that would be an uncharacteristically fast incompatible switch for Amazon, thus I doubt this somewhat. I bet an HVM image registered without sriov will still work, one will just not get deployed to the HW that supports and expects sriov. Anyhow without such a test this code is not OK to be merged. Also if it indeed turns out that sriov is required for HVM now then the IMHO nice approach to indicate this is to remove the option from the command line and document the behavior rather than magically hiding the details from the user. While pretending the user has a choice by retaining the command line option.

jgleissner:
Ok, understood. From reading the docs I thought sriov is simply mandatory for HVM, but then uploading an HVM image without the sriov switch used to work, so something must have changed on Amazon side. I'll investigate.

jgleissner:
I think I misunderstood the documentation. Apparently "there is no way to disable enhanced networking" means that you can't explicitly disable it using that parameter, but only by omitting it. I tested registering images using boto3 with and without sriov in register_image() and the result is as expected (i.e. image with and without sriov attrtibute). I'm not sure what changed but I'm guessing botocore used to accept SriovNetSupport=None as unset but now it rejects it.

I'll fix the register_image() call and open a new pull request.

rjschwei:
On 06/15/2016 06:44 AM, jgleissner wrote:

> I think I misunderstood the documentation. Apparently "there is no way to disable enhanced networking" means that you can't explicitly disable it using that parameter, but only by omitting it. I tested registering images using boto3 with and without sriov in register_image() and the result is as expected (i.e. image with and without sriov attrtibute). I'm not sure what changed but I'm guessing botocore used to accept SriovNetSupport=None as unset but now it rejects it.

Actually the "SriovNetSupport=None" is a left over from boto. When I
ported the utilities to boto3 the higher level objects that might handle
these things were not to the point where I could consistently use them.
Thus I decided to fall back to the lower order APIs and thus we have to
do more work with checking the arguments and issuing the proper commands.

This one just slipped through.

PR #90, opened 2016-06-15
ec2uploadimg: fix SRIOV flag (bsc#981224)

jgleissner:
Do not use SriovNetSupport parameter for register_image() if flag is
not set.

### DIFF
--- a/ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
+++ b/ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
@@ -696,15 +696,25 @@ class EC2ImageUploader(EC2Utils):
 
         root_device_name = self._determine_root_device()
         if self.image_virt_type == 'hvm':
-            ami = self._connect().register_image(
-                Name=self.image_name,
-                Description=self.image_description,
-                Architecture=self.image_arch,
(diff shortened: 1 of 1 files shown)


ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
```diff
@@ -696,15 +696,25 @@ def _register_image(self, snapshot):
...
+                )
+            else:
+                ami = self._connect().register_image(
+                    Name=self.image_name,
+                    Description=self.image_description,
+                    Architecture=self.image_arch,
+                    RootDeviceName=root_device_name,
+                    BlockDeviceMappings=block_device_map,
+                    VirtualizationType=self.image_virt_type
+                )
```
rjschwei, line 717: Rather than this implementation I think it would be construct a dictionary that is then used as kwargs dict, i.e. passed with **kwargs and populate the dictionary accordingly. Constructing the kwargs allows us not to have "sub-types" of the registration call for HVM images. We could also expand this concept to the PV images and thus only end up with one self._connect().register_image(...) call that simply receives the kwargs dict.

rjschwei:
Superseded by https://github.com/SUSE/Enceladus/pull/91

PR #93, opened 2016-08-17
ec2metadata: do not use proxy

jgleissner:
Ignore proxy settings when fetching data from EC2 meta data server.


rjschwei:
Code looks good. please bump the version in the spec file, we do not have many code changes for this utility and thus we want to get this change out the door. Also please file a bug for SLES and submit a new package, remember, package is maintained in OBS.

Thanks

PR #94, opened 2016-09-11
- ec2utils

rjschwei:
- Properly collect the Python packages into a namespace such that
  coalescing the individual source trees works properly at runtime.


moio:
I am no Python expert but I can confirm that the following now works as expected:

```
cd Enceladus/ec2utils/ec2utilsbase
sudo python setup.py install
cd ../ec2uploadimg/
sudo python setup.py install
ec2uploadimg --help
```

Thanks!

PR #100, opened 2016-10-04
Exit process when ssh key file not found.

### LINKED ISSUE #99: SSH Key Files Not Found
Edit: When uploading an image using ec2uploadimg:

If a user enters the incorrect location to an SSH key file there is an error message displayed but the upload process continues until attempting the SSH connection.

It seems like it would be useful to terminate the process along with the error message, similar to when the [ssh key file is not provided](https://github.com/SUSE/Enceladus/blob/master/ec2utils/ec2uploadimg/ec2uploadimg#L381).

smarlowucf:
If ssh key file is not found at the given location,
ssh_private_key_file, exit the upload process after
displaying an error message.

Fixed #99


rjschwei:
Thanks

PR #101, opened 2016-10-26
- ec2utils -- ec2uploadimg

rjschwei:
- Support enabling ENA device support for uploaded image.
- Version 2.0, incompatible change argument order


bear454:
Code looks fine, but I didn't thing `argparse` cares what order the arguments are listed in. If I'm right, and you're just adding an argument, that doesn't _need_ to be a major version bump (although it doesn't really hurt anything).

Code looks fine... lot of work just to set another flag ;-)

PR #103, opened 2016-11-18
ec2uploadimg: added support for using running instance

jgleissner:
I added support for using an existing instance for creating the snapshot instead of spinning up a new instance. This helps work around network issues like ssh connection to a newly created instance dies after a couple of minutes. ;) It could also speed up uploading multiple images since you wouldn't have to wait for instance creation for every one of them.

I also fixed --version, which threw an error.


ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
```diff
@@ -629,8 +631,12 @@ def _get_next_disk_id(self):
 
     # ---------------------------------------------------------------------
     def _launch_helper_instance(self):
-        """Launch the helper instance that is used to create the new image"""
         self._set_zone_to_use()
+        """When using an already running instance, simply look up this one"""
```
rjschwei: This logic should not be here. the method has a purpose, tat purpose is to launch the helper instance and thus it is named accordingly..

2 options:
Put the logic around where _launch_helper_instance is called i.e.
if self.running_ID:
    helper_instance = self._get_instance(self.running_ID)
else:
    helper_instance = self._launch_helper_instance()

or implement a generic implementation

def _get_helper_instance(self):
     if self.running_id:
         return ......
     else:
         return self._launch_helper_instance()

The current implementation introduces a method that has multiple purposes, look up or creation.

ec2utils/ec2uploadimg/python-ec2uploadimg.spec
```diff
@@ -18,7 +18,7 @@
 
 %define upstream_name ec2uploadimg
 Name:           python-ec2uploadimg
-Version:        2.0.0
+Version:        2.1.0
```
rjschwei: We use semantic versioning, you changed the interface to EC2ImageUploader thus this will result in 3.0.0

rjschwei (changes requested):

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
jgleissner: Fair point, I'll rearrange the code.

ec2utils/ec2uploadimg/python-ec2uploadimg.spec
jgleissner: semver.org reads "increment the minor version when you add functionality in a backwards-compatible manner". Which I believe I did.

rjschwei: Compatible would be if you add running_id at the end of the argument list to EC2ImageUploader, inserting it in the middle makes it an incompatible change. Consider if there is a developer that uses the EC2ImageUploader class and does not use keywork arguments, then what was previously passed as secret_key is now passed as running_id.

jgleissner: Ah right, you could use positional arguments only when instancing EC2ImageUploader, if you're a bit masochistic. ;) Didn't think of that, sorry. Bumping the major version for such a minor feature seems odd. Should I move running_id to the end of the argument list? It would break alphabetical order though.

rjschwei: PLease bump the version I like the alpha order

jgleissner:
I bumped the version and moved instance lookup into a separate function. Hope it's ok now.

rjschwei (approved):

PR #105, opened 2016-11-22
- regionServiceClient

rjschwei:
+ Implement region portability
    When a new image is created form a running instance the repositories and
    registration data a carried in the image. If this image is moved to
    another region the instance created in the new region does not register
    itself with the region local SMT servers. Resulting in either updates
    being pulled across regions, or no updates being available at all. This
    update addresses this problem by recognizing that the SMT data is
    different and taking appropriate action
  + This update also fixes a previously unknown hidden problem in that a new
    SMT server added to a region would not be picked up and cached.


regionServiceClient/lib/cloudregister/registerutils.py
```diff
@@ -106,6 +143,90 @@ def exec_subprocess(cmd):
...
+            logging.error('Unable to obtain SMT server information, exiting')
+            sys.exit(1)
+        smt_info = json.loads(response.text)
+        expected_entries = ('fingerprint', 'SMTserverIP', 'SMTserverName')
+        smt_info_xml = '<regionSMTdata><smtInfo '
+        for attr in expected_entries:
+            value = smt_info.get(attr)
+            if not value:
+                logging.error(
+                    'Metadata server did not supply a vlaue for "%s"' % attr
```
bear454: typo in "value"

regionServiceClient/lib/cloudregister/registerutils.py
```diff
@@ -106,6 +143,90 @@ def exec_subprocess(cmd):
 
 
 # ----------------------------------------------------------------------------
+def fetch_smt_data(cfg, proxies):
+    """Retrieve the data for the region SMT servers from a remote host"""
+    if cfg.has_option('server', 'metadata_server'):
+        metadata_url = cfg.get('server', 'metadata_server')
+        logging.info(
+            'Using metada server "%s" to obtain SMT information' % metadata_url
```
bear454: typo in "metadata"

regionServiceClient/usr/sbin/registercloudguest
```diff
@@ -187,146 +138,99 @@ if proxy:
...
+        if new_target:
+            msg = 'Configured SMT unresponsive, switching to equivalent '
+            msg += 'SMT server with ip %s' % new_target.get_ip()
+            utils.replace_hosts_entry(registration_smt, new_target)
+            utils.store_smt_data(
+                registration_smt_cache_file_name,
+                registration_smt
+            )
+        else:
+            msg = 'Configured SMT unresponsive, could not find '
```
bear454, line 223: Consider rewording this message.

bear454 (dismissed):

bear454:
Moving the code from the executable to utility in the same commit as the changes makes for a challenging review. Next time please move the code in a separate commit, then make changes.

PR #107, opened 2016-12-07
- regionServiceClient

rjschwei:
+ If the cache directory does not exist create it. Previously the assumption
    was made that the directory would exist, if it did not the registration
    failed


rjschwei:
@bear454 if you have a minute I'd appreciate if you could double check this

bear454 (commented):
:+1:

PR #117, opened 2017-02-10
- regionServiceClient

rjschwei:
+ Fix list name, failover code path used old variable name prior
    to implementing region portability


bear454 (approved):
I should have caught that before :/

PR #123, opened 2017-03-07
region hint

rjschwei:
Latent bug fix for the GCE plugin, implement tests, implement plugin for EC2

### DIFF
--- a/regionServiceClient/cloud-regionsrv-client.spec
+++ b/regionServiceClient/cloud-regionsrv-client.spec
@@ -80,6 +80,16 @@ Requires:     cloud-regionsrv-client >= 6.0.0
 %description plugin-gce
 Guest registration plugin for images intended for Google Compute Engine
 
+%package plugin-ec2
+Version:      1.0.0
+License:      LGPL-3.0
+Summary:      Cloud Environment Guest Registration Plugin for GCE
+Group:        Productivity/Networking/Web/Servers
+Requires:     cloud-regionsrv-client >= 6.0.0
+
+%description plugin-ec2
+Guest registration plugin for images intended for Amazon EC2
+
 %prep
 %setup -q
 
@@ -152,6 +162,9 @@ rm -rf $RPM_BUILD_ROOT
 %attr(0755,root,root) %{_initddir}/guestregister
 %{_sbindir}/rcguestregister
 %endif
+%dir %{python_sitelib}/cloudregister-%{base_version}-py%{py_ver}.egg-info
+%dir %{python_sitelib}/cloudregister/
+%{python_sitelib}/cloudregister-%{base_version}-py%{py_ver}.egg-info/*
 
 %files generic-config
 %defattr(-,root,root,-)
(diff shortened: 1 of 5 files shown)


regionServiceClient/cloud-regionsrv-client.spec
```diff
@@ -80,6 +80,16 @@ Requires:     cloud-regionsrv-client >= 6.0.0
 %description plugin-gce
 Guest registration plugin for images intended for Google Compute Engine
 
+%package plugin-ec2
+Version:      1.0.0
+License:      LGPL-3.0
+Summary:      Cloud Environment Guest Registration Plugin for GCE
+Group:        Productivity/Networking/Web/Servers
+Requires:     cloud-regionsrv-client >= 6.0.0
```
bear454: What's magical about 6.0.0 ?

regionServiceClient/lib/cloudregister/amazonec2.py
```diff
@@ -0,0 +1,40 @@
+# Copyright (c) $2017, SUSE LLC, All rights reserved.
```
bear454: What's the $ for ?

regionServiceClient/lib/cloudregister/amazonec2.py
```diff
@@ -0,0 +1,40 @@
...
+import logging
+import requests
+
+
+def generateRegionSrvArgs():
+    """
+    Generate arguments to be sent to the region server.
+    """
+    metaDataUrl = 'http://169.254.169.254/latest/meta-data/'
+    zoneInfo = 'placement/availability-zone'
```
bear454, line 23: What's the value in separating these?

regionServiceClient/lib/cloudregister/amazonec2.py
```diff
@@ -0,0 +1,40 @@
...
+
+    try:
+        zoneResp = requests.get(metaDataUrl + zoneInfo)
+    except:
+        msg = 'Unable to determine instance placement from "%s"'
+        logging.warning(msg % (metaDataUrl + zoneInfo))
+        return
+
+    if zoneResp.status_code == 200:
+        region = zoneResp.text[:-1]
```
bear454: I assume you're stripping a newline off the end of the response. I prefer `.strip()` to `[:-1]` both for readability, and ensuring what you drop really is whitespace.

regionServiceClient/tests/test_gceplugin.py
```diff
@@ -0,0 +1,111 @@
+# Copyright (c) $2017, SUSE LLC, All rights reserved.
```
bear454, line 1: Again with $2017 ?

regionServiceClient/cloud-regionsrv-client.spec
rjschwei: 6.0.0 is probably the first version that has plugin support

regionServiceClient/lib/cloudregister/amazonec2.py
rjschwei: Nope, we get the availability zone rather than the region. Obviously I need a comment. For example we get back

us-east-1f but we need us-east-1

Yes the implementation will fail when Amazon reaches more than 26 AZs per region. Lets cross that bridge when we get there.

PR #126, opened 2017-04-14
pint: add filter for 'type' attribute

bear454:


susePublicCloudInfoClient/setup.py
```diff
@@ -9,17 +9,18 @@
...
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 # GNU General Public License for more details.
-# 
+#
 # You should have received a copy of the GNU General Public License
 # along with ec2uploadimg. If not, see <http://www.gnu.org/licenses/>.
 
 import os
 import sys
+import version
```
rjschwei: This will surely fail. The import was in the previous location so we'd get the proper version file and not some version somewhere else

rjschwei (changes requested):

rjschwei:
Also I would have expected a test to get smt servers without filter that hen verifies that both smt-sles and smt-sap types are present. Did I miss that?

susePublicCloudInfoClient/setup.py
bear454: Putting it back...

```
$ flake8
./setup.py:35:1: E402 module level import not at top of file
```

bear454:
I think it's safe to assume that from the server tests.  The request URL isn't changing, and we're not doing any client-side manipulation outside the filters. I can add a test for it if you insist, but it feels redundant.

susePublicCloudInfoClient/setup.py
smarlowucf: What about an explicit execution of the version module?

```
this_path = os.path.dirname(os.path.abspath(__file__))
version_path = os.path.join(this_path, 'lib', 'susepubliccloudinfoclient', 'version.py')
execfile(version_path)

# In setup ...
version=VERSION
```

That way the python path is not being messed with, the join will use proper os.sep and flake8 remains happy.

rjschwei (approved):

susePublicCloudInfoClient/setup.py
bear454: @smarlowucf : would you mind submitting that in a separate PR ?

PR #127, opened 2017-04-14
Update readme docs.

smarlowucf:
- Add a section for installation on SUSE.
- Add long description.
- Add reference to man page for more info.

### DIFF
--- a/ec2utils/README.md
+++ b/ec2utils/README.md
@@ -1,21 +1,29 @@
-ec2utils
-========
+# ec2utils
 
 A collection of utilities that makes managing images in Amazon EC2 easier.
 All utilities utilize the same configuration file, ~/.ec2utils.conf that
 contains account and region configuration options.
 
-# ec2deprecateimg
+## ec2deprecateimg
 
-A tool to deprecate images. Images are tagged with:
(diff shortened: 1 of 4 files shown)


rjschwei:
Why?

The agreement is that our internal doc is asciidoc, this is external documentation that gets delivered with the package. Which implies that a change to the file name would also have to be reflected in the .spec file.

smarlowucf:
Two parts:
- Adding info to the readme files since they were empty aside from one line description.
- Chose asciidoc as I thought this was preferred. (forgot to update spec :disappointed:)

These can be converted back to markdown if the updates to the readme files are agreeable.

rjschwei:
All ec2utils come with man pages, we should not repeat a good chunk of the man page information in the README as that implies when changes occur we have to maintain 2 places for documentation. That also implies that at least 1 place will be out of date.

So we should go back to .md format and the content change should be much smaller with a reference to the man page.

smarlowucf:
Removed the options from each readme and converted back to md.

I think it's useful to keep the long descriptions for those who come across the utilities via Github. And similarly providing a small installation section specific for now to SUSE.

PR #129, opened 2017-04-17
Use exec builtin to source version number from file.

smarlowucf:
From #126, cleanup flake8 error in version import.

### DIFF
--- a/susePublicCloudInfoClient/setup.py
+++ b/susePublicCloudInfoClient/setup.py
@@ -27,13 +27,15 @@ except ImportError:
     sys.stderr.write('Python setuptools required, please install.')
     sys.exit(1)
 
-
-this_path = os.path.dirname(os.path.abspath(__file__))
-mod_path = this_path + os.sep + 'lib/susepubliccloudinfoclient'
-sys.path.insert(0, mod_path)
-
-import version
-src_version = version.VERSION
+version = {}
+version_path = os.path.join(
+    os.path.dirname(os.path.abspath(__file__)),
+    'lib',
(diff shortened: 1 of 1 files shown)


rjschwei:
And this is better why?

No, flake8 complains is not the proper answer

bear454:
I asked @smarlowucf to submit this suggestion as a separate PR... it is better, if we're using flake8, to have a clean flake8 report... so is there some _problem_ with this approach @rjschwei ?

smarlowucf:
Explicit execution on module rather than implicit import relying on a modified python path.

May not be likely but the python path could theoretically change between modifying it and running import. 

```
sys.path.insert(0, mod_path)
# Something manages to change python path
import version
```

The downfall to a module execution is it will always be executed. Granted this module is negligible with one string.

smarlowucf:
Another option would be to use bumpversion to manage the version in multiple places. Moving the logic outside of the setup file.

bear454:
:+1: for moving to `bumpversion`

rjschwei:
OK, I get the argument about not getting used to warnings, fair enough. But why is there a tendency to find the complicated solution?

import lib.susepubliccloudinfoclient.version
from lib.susepubliccloudinfoclient.version import *

yes it is a relative import, but it is in setup.py and not part of the code. So we can appease flake8 and still keep things simple.

rjschwei:
OK, so we are not going to do this!

exec => means as an attacker I can slip whatever code I want into the version file and then have it executed as root during package installation, same holds true for "import" of course. We should move to a "read and extract" method without having any potential code in the version file executed.

PR #142, opened 2017-08-15
- gcemetadata

rjschwei:
+ Properly handle identity requests. The identity query needs an argument
    to generate a JWT, to this point the information for --identity was
    always displayed as "unavailable". Now the command line option takes
    an argument that is passed to the API to generate a JWT


jgleissner (approved):
Looks ok

smarlowucf (approved):

PR #146, opened 2017-08-17
- gcemetadata

rjschwei:
+ Support setting the format for the identity token, supported
    values are "standard" or "full" per the API spec.


jgleissner (approved):

gcemetadata/gcemetadata
```diff
@@ -115,9 +115,7 @@ def main():
         getopt_metaopts.append('diskid=')
         getopt_metaopts.append('licenseid=')
         getopt_metaopts.append('netid=')
-        getopt_metaopts.append('output=')
         getopt_metaopts.append('query=')
```
smarlowucf, line 118: Did these accidentally get removed? (output, and xml)

gcemetadata/gcemetadata
```diff
@@ -180,6 +178,14 @@ def main():
...
                 sys.exit(1)
             meta.set_identity_arg(val)
+        elif opt == '--identity-format':
+            supported_vals = ['full', 'standard']
+            if val not in supported_vals:
+                msg = '--identity-format possible values: "%s"' %supported_vals
+                print >> sys.stderr, msg
+                sys.exit(1)
+            meta.set_identity_format(val)
+            continue
```
smarlowucf, line 188: Unnecessary continue

gcemetadata/gcemetadata
rjschwei: No they are in the general options list, which gets added, thus these were duplicate entries all along

gcemetadata/gcemetadata
rjschwei: No the continue is necessary of the --identity-format which is not an endpoint url will be added to the options passed for query and thus "unavailable" will be printed.

PR #148, opened 2017-08-23
- gcemetadata

rjschwei:
+ pep8 compliance
  + Modify function, method, and variable names

### DIFF
--- a/gcemetadata/gcemetadata
+++ b/gcemetadata/gcemetadata
@@ -25,39 +25,43 @@ import gcemetadata.gcemetautils as gcemetautils
 
 from gcemetadata.gcemetaExceptions import *
 
-generalOptions = ['api', 'help', 'identity-format=', 'listapis', 'output=',
-                  'version', 'xml']
+general_options = ['api', 'help', 'identity-format=', 'listapis', 'output=',
+                   'version', 'xml']
 opts_with_args = ['identity']
-opts_with_id_targets = ['disks','licenses', 'network-interfaces']
+opts_with_id_targets = ['disks', 'licenses', 'network-interfaces']
 
-def printOption(level, indent, options):
+
+def print_option(level, indent, options):
     prefix = '--'
     for opt in options.keys():
         if type(options[opt]) == dict:
             if opt in opts_with_id_targets:
                 # Figure out the ID
-                devIDs = options[opt].keys()
-                devIDs.sort()
-                for devID in devIDs:
+                option_ids = options[opt].keys()
+                option_ids.sort()
+                for option_id in option_ids:
                     print >> sys.stdout, indent * level + prefix + opt
                     if opt == 'disks':
-                        info = indent * (level + 1) + prefix + 'diskid ' + devID
+                        info = indent * (level + 1)
+                        info += prefix + 'diskid ' + option_id
                         print >> sys.stdout, info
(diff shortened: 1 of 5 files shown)


gcemetadata/gcemetadata
```diff
@@ -25,40 +25,44 @@ import gcemetadata.gcemetautils as gcemetautils
...
     for opt in options.keys():
         if type(options[opt]) == dict:
             if opt in opts_with_id_targets:
                 # Figure out the ID
-                devIDs = options[opt].keys()
-                devIDs.sort()
-                for devID in devIDs:
+                option_ids = options[opt].keys()
+                option_ids.sort()
+                for opt_id in option_ids:
```
bear454: I'd prefer the singular match the plural; don't see any good reason to shorten it. Consider:

```
for option_id in option_ids:
```

gcemetadata/lib/gcemetadata/gcemetadata.py
```diff
@@ -21,47 +21,44 @@
 
 from gcemetaExceptions import *
```
smarlowucf, line 23: Ditto

gcemetadata/gcemetadata
```diff
@@ -25,40 +25,44 @@ import gcemetadata.gcemetautils as gcemetautils
 
 from gcemetadata.gcemetaExceptions import *
```
smarlowucf, line 26: Wildcard imports should be avoided.

gcemetadata/lib/gcemetadata/gcemetautils.py
```diff
@@ -26,43 +26,46 @@
 
 from gcemetaExceptions import *
```
smarlowucf, line 27: Ditto

smarlowucf (changes requested):
Looks good, a few other things:

Pep8:
- setup.py:40 remove whitepspace before colon
- gcemetaExceptions.py:18 expected 2 blank lines before class
- gcemetadata.py:69 `e` is never used

Syntax:
- gcemetadata.py:70 time is undefined
- gcemetautils.py:75 s/filePath/path
- setup.py:24-25 sys is undefined

gcemetadata/gcemetadata
rjschwei: Done

gcemetadata/lib/gcemetadata/gcemetautils.py
rjschwei: Sorry, in context I don't see where "import *" does any harm in potentially confusing name spaces, thus not really seeing a need for a change.

Everything else is done

smarlowucf: Agreed, but it is mentioned as part of [pep8](https://www.python.org/dev/peps/pep-0008/#imports) even if `pep8` doesn't raise a warning. The `flake8` wrapper however does raise exception for all instances of wildcard usage + usage of imports that aren't explicitly imported.

E.g.
`F405 'GCEMetadataException' may be undefined, or defined from star imports: gcemetaExceptions`
`F403 'from gcemetaExceptions import *' used; unable to detect undefined names`

smarlowucf (approved):

bear454 (approved):

PR #150, opened 2017-08-25
Remove egg info from source

smarlowucf:


bear454:
Why?

smarlowucf:
Just cleanup, the egg info really shouldn't be committed to source.

PR #151, opened 2017-09-06
- monitor

rjschwei:
+ Implement new plugin to verify if a directory exists and if it does
    check whether or not it is empty. This is needed to watch if
    registration sharing is failing and generating backlogs.

### DIFF
--- a/monitor/check_dir_empty
+++ b/monitor/check_dir_empty
@@ -1,0 +1,59 @@
+#!/usr/bin/python3
+
+# Copyright (c) 2017 SUSE LLC  All rights reserved.
+#
+# check_dir_empty is free software: you can redistribute it
+# and/or modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation, version 2 of
+# the License.
+#
+# check_dir_empty is distributed in the hope that it will
+# be useful, but WITHOUT ANY WARRANTY; without even the implied warranty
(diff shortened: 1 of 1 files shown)


monitor/check_dir_empty
```diff
@@ -0,0 +1,59 @@
...
+if not dir_to_monitor:
+    print(__doc__)
+    sys.exit(WARNING)
+
+if not os.path.exists(dir_to_monitor):
+    sys.exit(OK)
+else:
+    content = glob.glob('%s/*' % dir_to_monitor)
+    if content:
+        print('The directory "%s" does exist an is not empty' % dir_to_monitor)
```
smarlowucf: s/an/and

monitor/check_dir_empty
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/python3
+
+# Copyright (c) 2017 SUSE LLC  All rights reserved.
+#
+# check_dir_empty is free software: you can redistribute it
+# and/or modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation, version 2 of
```
smarlowucf, line 7: Should this be version 3?

smarlowucf (changes requested):

monitor/check_dir_empty
rjschwei: Nope, needs to be v2 for nagios compatibility

smarlowucf (approved):

PR #153, opened 2017-09-12
Features and bug fix

rjschwei:
- Fix python 3 conversion bug, exceptions no longer have a .message member, use format() instead
- Support adding the billing product code during registration
- Read the VPC subnet ID from the config if not specified


bear454:
On the comment of 0e8859bd, `get he message` should be `get the message`.

Other than that, looks great.

bear454 (approved):

ec2utils/ec2uploadimg/lib/ec2utils/upload_VERSION
```diff
@@ -1 +1 @@
-5.0.0
+5.0.1
```
bear454: Consider having this be a minor bump, as you're adding a new feature (specifying VPC subnet in config files).

ec2utils/ec2utils.conf.example
```diff
@@ -11,6 +11,7 @@ access_key_id =
 secret_access_key = 
 ssh_key_name =
 ssh_private_key =
+subnet_id_us-east-1 = subnet-123456
```
smarlowucf, line 14: Could subnet_id be in region section?

```
[region-us-east-1]
subnet_id = subnet-123456
```

smarlowucf (approved):

ec2utils/ec2uploadimg/lib/ec2utils/upload_VERSION
rjschwei: The major bump is from inserting the billing_codes argument into the ctor of the imaplementation which makes the API incompatible

ec2utils/ec2utils.conf.example
rjschwei: subnet is region and account specific, for example I upload with -a rschweikert and -a resale-images to us-east-1, the subnet cannot be shared between these accounts so it needs to be account specific, but also needs to be region specific because subnets do not span regions.

Concocting the config value name from "fixed-identifier" + region was from my point of view the most straight forward option otherwise the config file might have needed to be re-constructed, which would be more painful, IMHO

PR #157, opened 2017-10-06
- regionServiceClient

rjschwei:
+ Port to Python 3


regionServiceClient/lib/cloudregister/registerutils.py
```diff
@@ -348,7 +348,7 @@ def get_smt_from_store(smt_store_file_path):
         return None
 
     smt = None
-    with open(smt_store_file_path, 'r') as smt_file:
+    with open(smt_store_file_path, 'rb') as smt_file:
```
bear454, line 351: Did the content format change from text to binary?

smarlowucf (approved):

regionServiceClient/lib/cloudregister/registerutils.py
rjschwei: It's the py3 bytes vs string thingy when reading a picked object.....

PR #159, opened 2017-10-11
Cloud netconfig fixes

jgleissner:
Use Azure metadata API version 2017-08-01 (2017-07-01 is not available anymore)
Handle potentially slow metadata update in EC2


cloud-netconfig/azure/functions.cloud-netconfig
```diff
@@ -18,7 +18,7 @@
 
 METADATA_BASE_URL="http://169.254.169.254/metadata/instance/network/interface/"
 URL_HDR="Metadata:true"
-URL_APX='?format=text&api-version=2017-07-01'
+URL_APX='?format=text&api-version=2017-08-01'
```
rjschwei, line 21: There should be a way to get the available versions for the API. Given it is versioned it should never go away but as this change suggests Msft may do exactly that and remove a versioned API, go figure.

Given that we are getting under time pressure this is OK for now, however I created https://github.com/SUSE/Enceladus/issues/160 to ensure we come back to this.

rjschwei (approved):

cloud-netconfig/azure/functions.cloud-netconfig
jgleissner: I was surprised too, but it really seems they removed that API version. Perhaps it was never official, but that's hard to tell. I'll look into a lookup mechanism.

PR #162, opened 2017-10-13
cloud-netconfig: use tested and supported metadata API versions

jgleissner:
Explicitely use API version 2016-09-02 in EC2
Use API version 2017-04-02 (latest version that is available in all regions)


jgleissner:
Forgot to bump version, I'll push another commit.

rjschwei (approved):

PR #163, opened 2017-11-16
- ec2utils

rjschwei:
+Prepare for submission to pypi

### DIFF
--- a/ec2utils/ec2utilsbase/Makefile
+++ b/ec2utils/ec2utilsbase/Makefile
@@ -31,5 +31,13 @@ tar: clean
 test:
 	py.test tests/ec2utilsutilstest.py
 
+pypi:
+	mkdir -p "$(NAME)-$(verSrc)"/man/man1
+	cp -r $(dirs) $(files) "$(NAME)-$(verSrc)"
+	tar -czf "$(NAME)-$(verSrc).tar.gz" "$(NAME)-$(verSrc)"
+	rm -rf "$(NAME)-$(verSrc)"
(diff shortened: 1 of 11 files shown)


schaefi (approved):
Looks great 👍

ec2utils/ec2utilsbase/Makefile
```diff
@@ -31,5 +31,13 @@ tar: clean
 test:
 	py.test tests/ec2utilsutilstest.py
 
+pypi:
```
smarlowucf, line 34: Sorry don't remember getting pinged about this review. Anyway, looks good. Although, it seems a bit redundant to build sdist in makefile when python setup.py already provides this functionality.

rjschwei: sdist leaves stuff out, such as the man pages for example. If we put this on pypi then other packagers should get the complete source and not just what the Python tools consider important to Python.

Yes I tried playing around with additional settings in setup.py to include the man pages, but that was unreliable and I am not interested in changing the way the man pages are written just because the Python tools cannot cope properly.

smarlowucf: Yeah, it is confusing what flags do the trick given changes between distutils and setuptools. With IPA using [`include_package_data`](https://github.com/SUSE/ipa/blob/master/setup.py#L74) and recursive include of man in [MANIFEST](https://github.com/SUSE/ipa/blob/master/MANIFEST.in) seemed to do the trick.

PR #165, opened 2017-11-30
Fixing namespace issues when using module from pip

schaefi:
When installing ec2utils from pip none of the tools are working
because of invalid imports. The way the code is packed on pip
does not work with the way the modules are imported in the
code. This patch makes ec2utils to work when installed from
pip. It also fixes the use of print to be compatible with
python2 vs. python3


schaefi:
Please Note:

I stumbled over all those issues when testing the modules from pip. In order to do that I have uploaded them to my personal account. We can delete it from there if needed at a later point in time, just to let you know the pip index now has ec2utilsbase and ec2uploadimg reserved

Important to note is that I think this patch fixes the module when used from pip but I think it will break the code when it is installed from the rpm package. I'm not sure because after hours of trying to fix the namespace issues I ran out of energy to check how the code is installed through packages. Maybe it's not an issue at all but if the ec2utils rpm packages are installed differently than the the sdist target on pip I strongly recommend to not do so and keep the same structure no matter how the code is packaged

Therefore before this is merged one needs to check if the rpm packages build with this patch still works. Overall I recognized that the structure of the source in Enceladus/ec2utils does not follow the suggested structure for python modules. The most problematic part is that you use source files named the same than source directories. e.g ec2utilsbase/lib/ec2utils/ec2utils.py
This causes many headaches in python on import

Let's see if this can be used as api and tool without too many changes

So again the patches here were only tested with the module installed from pip

schaefi:
I also fixed setup.py at various places which is required to successfully run

```
python setup.py sdist upload -r pypi
```

schaefi:
Related to this changes the use of ec2uploadimg was activated here: https://github.com/SUSE/mash/pull/30

ec2utils/ec2utilsbase/setup.py
```diff
@@ -46,11 +46,11 @@
             '': 'lib',
         },
         namespace_packages = ['ec2utils'],
-        classifiers=(
+        classifiers=[
             'Development Status :: 5 - Production/Stable',
             'Intended Audience :: Developers',
             'Natural Language :: English',
-            'License :: OSI Approved :: GPL-3.0+',
+            'License :: OSI Approved :: GNU General Public License v3 (GPLv3)',
```
smarlowucf: License classifier should be:

License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
```diff
@@ -23,8 +24,8 @@
 import time
 
 
-from ec2utils.ec2utils import EC2Utils
-from ec2utils.ec2UtilsExceptions import *
+from ec2utils import EC2Utils
```
smarlowucf: Relative imports break Python3 compatibility.

smarlowucf (changes requested):

smarlowucf:
> python setup.py sdist upload -r pypi

Should use twine for upload instead of setup.py. https://packaging.python.org/tutorials/distributing-packages/#uploading-your-project-to-pypi

smarlowucf:
> I stumbled over all those issues when testing the modules from pip. In order to do that I have uploaded them to my personal account. We can delete it from there if needed at a later point in time, just to let you know the pip index now has ec2utilsbase and ec2uploadimg reserved

@schaefi If you go to the old PyPI you can just add new owners/maintainers to the packages. https://pypi.python.org/pypi

schaefi:
Tested both modules from pip for py2 and py3, worked so far

smarlowucf (approved):

smarlowucf:
Looks good, strange that python2 works with explicit relative imports but not absolute :confused:. Although, it appears to be related to the not recommended usage of wildcards.

rjschwei:
But ec2utils was specifically converted to be Python3 only. Not sure why we go backwards. I do not like this change. Sorry I was at a conference and had no time to review this earlier. Given that there is really no time pressure you could have waited with the merge.

rjschwei:
I added the make pip target to let us upload with twine and not have to deal with uploading via setup.py because that's not what we should be doing.

rjschwei:
Also I added a team account to the pypi and the credentials to our secrets repo for a reason.

schaefi:
>    But ec2utils was specifically converted to be Python3 only. Not sure
>    why we go backwards.

py2 is not dead, all the boto code you use works in py2 too, making
your own code to work with py2 was two lines of code. The rest is
cleanup (no glob imports) and a compatible way to import. I don't
see it as a step back, especially for modules on pip

>    I do not like this change. Sorry I was at a
>    conference and had no time to review this earlier.

yes sorry about it, I should have waited with the merge

> Given that there is
> really no time pressure you could have waited with the merge.

I slightly disagree, you want mash to be operational at least for ec2
in March 2018. There are two contributors and many work to be done.
Counting the average speed, the current state and the task list I
provided in the card plus the fact that the X-mas vacation period
starts soon, I would say we have roughly 12Weeks of active development
and I did not took into account all the other work

So sorry, this needs going and I want the uploader service to be
_done_ before I go for vacation. So I need your stuff in pip and be
happy that it works there, see my comments in the card

But yes we could have waited until you are back

Sorry again

PR #167, opened 2017-12-14
Add --session-token parameter

ChrisBr:
which is necessary to authenticate with temporary security credentials.
http://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html.

### DIFF
--- a/ec2utils/ec2uploadimg/ec2uploadimg
+++ b/ec2utils/ec2uploadimg/ec2uploadimg
@@ -155,6 +155,12 @@ argparse.add_argument(
     metavar='AWS_SECURITY_GROUP_IDS'
 )
 argparse.add_argument(
+    '--session-token',
+    dest='sessionToken',
+    help='AWS session token, for use of temporary credentials (Optional)',
+    metavar='AWS_SESSION_TOKEN'
+)
+argparse.add_argument(
     '--snaponly',
     action='store_true',
     default=False,
@@ -515,6 +521,7 @@ try:
                           running_id=running_id,
                           secret_key=secret_key,
                           security_group_ids=args.securityGroupIds,
+                          session_token=args.sessionToken,
                           sriov_type=sriov_type,
                           ssh_key_pair_name=key_pair_name,
                           ssh_key_private_key_file=ssh_private_key_file,
--- a/ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
+++ b/ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
@@ -48,6 +48,7 @@ class EC2ImageUploader(EC2Utils):
                  running_id=None,
                  secret_key=None,
                  security_group_ids='',
+                 session_token=None,
                  sriov_type=None,
                  ssh_key_pair_name=None,
                  ssh_key_private_key_file=None,
@@ -75,6 +76,7 @@ class EC2ImageUploader(EC2Utils):
         self.running_id = running_id
         self.secret_key = secret_key
         self.security_group_ids = security_group_ids
+        self.session_token = session_token
         self.sriov_type = sriov_type
         self.ssh_key_pair_name = ssh_key_pair_name
         self.ssh_key_private_key_file = ssh_key_private_key_file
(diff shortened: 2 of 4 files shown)


ec2utils/ec2uploadimg/ec2uploadimg
```diff
@@ -145,6 +145,12 @@ argparse.add_argument(
     help='AWS secret access key (Optional)',
     metavar='AWS_SECRET_KEY'
 )
+argparse.add_argument(
+    '-o', '--session_token',
```
ChrisBr: @rjschwei not sure about the ``-o`` though. ``-s`` and ``-t`` is already in use.

ec2utils/ec2uploadimg/ec2uploadimg
```diff
@@ -145,6 +145,12 @@ argparse.add_argument(
     help='AWS secret access key (Optional)',
     metavar='AWS_SECRET_KEY'
 )
+argparse.add_argument(
+    '-o', '--session-token',
```
rjschwei: Either use "t" as the short name or remove the short option, "o" would make me think of "output"

ec2utils/ec2uploadimg/ec2uploadimg
```diff
@@ -514,6 +520,7 @@ try:
                           root_volume_size=root_volume_size,
                           running_id=running_id,
                           secret_key=secret_key,
+                          session_token=args.sessionToken,
```
rjschwei: Breaks alpha order should be "...secret_key,security_group_ids,session_token...".

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
```diff
@@ -47,6 +47,7 @@ def __init__(self,
                  root_volume_size=10,
                  running_id=None,
                  secret_key=None,
+                 session_token=None,
```
rjschwei: dito, alpha order

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
```diff
@@ -74,6 +75,7 @@ def __init__(self,
         self.root_volume_size = int(root_volume_size)
         self.running_id = running_id
         self.secret_key = secret_key
+        self.session_token = session_token
```
rjschwei: dito alpha order

ec2utils/ec2uploadimg/man/man1/ec2uploadimg.1
```diff
@@ -125,6 +125,8 @@ Specifies the AWS secret access key and overrides the value given for the
 with the
 .I secret_access_key
 in the configuration file.
+.IP "-o --session-token AWS_SESSION_TOKEN"
```
rjschwei: Short option, see comment above, and alphorder

ec2utils/ec2utilsbase/lib/ec2utils/ec2utils.py
```diff
@@ -39,6 +39,7 @@ def _connect(self):
             ec2 = boto3.client(
                 aws_access_key_id=self.access_key,
                 aws_secret_access_key=self.secret_key,
+                aws_session_token=self.session_token,
```
rjschwei: I suspect this will break, the boto client is not as forgiving as one would hope with accepting arguments that are "None" and then doing the "right" thing.

So we either need to run a test before taking this code or change the code to be conditional, i.e.

if self.session_token:
  ec2 = boto3.client( aws_session_token=self.session_token,...)
else:
  ec2 = boto3.client(aws_access_key_id=self.access_key,...)

rjschwei (changes requested):

ec2utils/ec2utilsbase/lib/ec2utils/ec2utils.py
ChrisBr: I tried it locally on my machine at it worked fine!

ec2utils/ec2uploadimg/ec2uploadimg
ChrisBr: -t is already used by type

ec2utils/ec2utilsbase/lib/ec2utils/ec2utils.py
ChrisBr: I changed it anyway to be sure ...

rjschwei (approved):

ChrisBr:
Thanks a lot @rjschwei

PR #169, opened 2017-12-21
Fix imports in ec2utilsbase

smarlowucf:
- Fix imports for ec2utilsbase. Using namespace means there's no init file in base package thus even explicit relative imports don't work.


rjschwei (changes requested):
see comments for ec2deprecateimg

rjschwei (approved):

PR #170, opened 2017-12-21
Fix imports in ec2publishimg

smarlowucf:
- Update gitignore to ignore packaging files and dirs.


rjschwei (changes requested):
see comments for ec2deprecateimg

rjschwei (approved):

PR #171, opened 2017-12-21
Fix imports in ec2uploadimg

smarlowucf:


rjschwei (changes requested):
See comments for ec2deprecateimg

rjschwei (approved):

PR #174, opened 2018-01-05
Implement targetFilename parameter

ChrisBr:
to make it possible renaming the source while uploading to the target.

The purpose of the feature is that for our uploader job the OBS backend renames the files to ``ID.file``, e.g. ``42.file``. ec2uploadimg can not handle these file extensions (and ec2uploadimg needs to know how handle the uploaded file, e.g. extracting). Therefore I propose to make it possible to hand over a filename and not only the source to rename the source while uploading to the target.

@rjschwei does this make sense?

### DIFF
--- a/ec2utils/ec2uploadimg/ec2uploadimg
+++ b/ec2utils/ec2uploadimg/ec2uploadimg
@@ -187,6 +187,11 @@ argparse.add_argument(
     type=int
 )
 argparse.add_argument(
+    '--target-filename',
+    dest="targetFilename",
+    help='The target filename'
+)
+argparse.add_argument(
     '-t', '--type',
     dest='instType',
     help='Instance type to use to upload image (Optional)',
@@ -541,7 +546,7 @@ try:
             ami = uploader.create_image_use_root_swap(args.source)
             print('Created image: ', ami)
         else:
-            ami = uploader.create_image(args.source)
+            ami = uploader.create_image(args.source, args.targetFilename)
             print('Created image: ', ami)
 except EC2UploadImgException as e:
     print(format(e), file=sys.stderr)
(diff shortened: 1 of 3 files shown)


rjschwei:
It appears to me that we are adding extra complications. the output of an image build via kiwi in OBS is a *.raw.xz file. If I understand your comment correctly this then gets renamed to ID.file and we then have to deal with what could be considered a superfluous rename at the source level.

Therefore my first question would be why do we rename, in OBS, a file that is perfectly suitable for uploading as is. I can understand renaming if there is a global image store and there would be potential naming conflicts for various builds if 2 users happen to give their images the same name, but builds are stored by project and thus no naming conflict should arise.

Sorry I don't understand why this is necessary.

ChrisBr:
Hi Robert,

> I can understand renaming if there is a global image store and there would be potential naming conflicts for various builds if 2 users happen to give their images the same name, but builds are stored by project and thus no naming conflict should arise.

Actually there is a global image store, that is the reason why we need to rename the images to avoid naming conflicts. The OBS itself is organized in projects & packages, however, for the upload the images get copied to another machine, the image upload worker which is the global image store. We could also recreate project and package folders but we figured this would be the easiest solution. Also a user could create several upload jobs for the same image, e.g. to different regions (which would happen in different jobs and the uploader would download the same image to the global image storage).

However, I can understand if you're completely against this feature. Then we have to go down the road and store everything in folders but we just figured this would be the simplest solution.

rjschwei:
Based on this it would appear to me that if the original rename would simply be to ID.raw.xz rather than ID.file the "rename on upload" would not be necessary while at the same time OBS can hold on to the flat storage space. In the end the user doesn't really care what the image source or the target name is, what the user cares about is the image name and that is hopefully provided in some setup UI for the upload.

ChrisBr:
Thanks Robert! I will close for now and talk with the backend guys!

PR #175, opened 2018-01-05
Fix find images by name functions to handle pending images

smarlowucf:
If image is pending there will be no Name key. Use get method to return default None if key does not exist.

- Update unit tests
- Add regression test to cover pending image with no name
- Remove unused os import
- Fix _no_name_warning function to accept dict not object


ec2utils/ec2utilsbase/lib/ec2utils/ec2utilsutils.py
```diff
@@ -75,10 +74,10 @@ def find_images_by_name(images, image_name):
     """Return a list of images that match the given name."""
     matching_images = []
     for image in images:
-        if not image['Name']:
+        if not image.get('Name'):
             print(_no_name_warning(image))
             continue
-        if image_name == image['Name']:
+        if image_name == image.get('Name'):
```
rjschwei: When the flow of control gets here we know the 'Name' key exists thus we can use "direct" access ['Name'] dito for other occurrences.

rjschwei (changes requested):

smarlowucf:
Mucked up rebase (with two older commits) in making changes will create clean PR.

PR #176, opened 2018-01-08
Fix find images by name functions to handle pending images

smarlowucf:
If image is pending there will be no Name key. Use get method to return default None if key does not exist.

- Update unit tests, fix test file name for pytest autodiscovery
- Add regression test to cover pending image with no name
- Remove unused os import
- Fix _no_name_warning function to accept dict not object

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@ var/
 *.egg-info/
 .installed.cfg
 *.egg
+
+# Unit test
+.cache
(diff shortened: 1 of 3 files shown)


smarlowucf:
Reverted to directly access dictionary 'Name' key when we know the key exists and is not None. Also, changed unit test file name so it will be discovered by pytest automagically.

rjschwei (approved):

PR #177, opened 2018-01-09
Get regions based on account partition.

smarlowucf:
Regions will return based on the account partition. This could be aws, aws-cn or aws-us-gov.

### DIFF
--- a/ec2utils/ec2utilsbase/lib/ec2utils/ec2utilsutils.py
+++ b/ec2utils/ec2utilsbase/lib/ec2utils/ec2utilsutils.py
@@ -181,10 +181,15 @@ def get_regions(command_args, access_key, secret_key):
     if command_args.regions:
         regions = command_args.regions.split(',')
     else:
-        regions = boto3.session.Session(
+        session = boto3.session.Session(
             aws_access_key_id=access_key,
             aws_secret_access_key=secret_key
-        ).get_available_regions('ec2')
+        )
+        client = session.client('sts')
+        partition = client.get_caller_identity()['Arn'].split(':')[1]
+        regions = session.get_available_regions(
+            'ec2', partition_name=partition
+        )
 
     return regions
 
--- a/ec2utils/ec2utilsbase/lib/ec2utils/ec2utilsutils.py
+++ b/ec2utils/ec2utilsbase/lib/ec2utils/ec2utilsutils.py
@@ -174,7 +174,7 @@ def get_from_config(account, config, region, entry, cmd_line_arg):
 
 
 # -----------------------------------------------------------------------------
-def get_regions(command_args, access_key, secret_key):
+def get_regions(command_args, access_key, secret_key, partition=None):
     """Return a list of connected regions if no regions are specified
        on the command line"""
     regions = None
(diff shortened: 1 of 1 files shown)


rjschwei:
I think I know what you are after but this does not have the desired effect. When creating a session with credentials valid for China:

>>> import boto3
>>> session = boto3.session.Session(aws_access_key_id=access_key,aws_secret_access_key=secret_key)
>>> client = session.client('sts')
>>> partition = client.get_caller_identity()['Arn'].split(':')[1]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/site-packages/botocore/client.py", line 314, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/usr/lib/python2.7/site-packages/botocore/client.py", line 612, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (InvalidClientTokenId) when calling the GetCallerIdentity operation: The security token included in the request is invalid.

Anyway, this would obviously break all the tools.

rjschwei (changes requested):
I'll leave it to you if you want to dig into the issue with China or scrap this and handle the issue at the next level up. Given that China has only 2 regions at this point this is still easy to handle in other ways.

rjschwei:
Last but not least not e that the API is unreliable, I have pointed that out to AWS previously but didn't have a good reproducer. With eu-west-3 only I can now also show that eu-west-3 is missing in an AWS instance in EC2. I have pointed the problem out to AWS, but we might want to rethink this in a different way, which would probably include using the config file somehow.

rjschwei:
And one more thing, the access of the partition when authenticated to gov-cloud also fails.

smarlowucf:
> And one more thing, the access of the partition when authenticated to gov-cloud also fails.

Ah yes, I had this working for both when testing but I realize the issue. If a specific profile or region is not passed in it's pulling the region from default profile. Thus it's trying to authenticate against a connected region. Will see if there's a way around this...

smarlowucf:
> I'll leave it to you if you want to dig into the issue with China or scrap this and handle the issue at the next level up. Given that China has only 2 regions at this point this is still easy to handle in other ways.

I have similar code the next level up but figured it would be helpful here as well. Will see if there's a possible workaround for default profile...

smarlowucf:
>Last but not least not e that the API is unreliable, I have pointed that out to AWS previously but didn't have a good reproducer. With eu-west-3 only I can now also show that eu-west-3 is missing in an AWS instance in EC2.

Had the same issue with the new China region, the problem was the version of botocore. The regions are [hardcoded](https://github.com/boto/botocore/blob/f68c0d0e386b585267a173a89268e7338b45151a/botocore/data/endpoints.json) thus every new region requires a new release of the package.

rjschwei:
Good find on the regions and here I was under the impression that there was finally an API, we had the problem with update python-boto for every new region, I guess that just shifted to botocore. I want to go throw up now.

smarlowucf:
> I want to go throw up now.

😃 It's already json why can't it just be provided by an API?

Anyway, closing this for now. I think the best fix would require a bit of refactoring in publish and deprecate scripts to handle aws profiles and/or another region arg.

We could also pull the region info directly from json which would remove the issue of package version. But this really should be part of the AWS API that boto relies on...

Will note to look into this further.

PR #178, opened 2018-01-09
Use validate method to validate share-with arg.

smarlowucf:
This is minor given it's an error case but any value thrown at the current regex yields an instance of _sre.SRE_Match. Thus it always evaluates to True.

Tried making a suitable regex but after fighting with it I chose to just use Python split with a separate validate function.

Given the use of split maybe this should be re-worked more to store the validated and split list so it doesn't have to be parsed twice. Also, the validate function could be added to the generic utils file instead of creating a new one for publish.

### DIFF
--- a/ec2utils/ec2publishimg/ec2publishimg
+++ b/ec2utils/ec2publishimg/ec2publishimg
@@ -17,10 +17,8 @@
 # You should have received a copy of the GNU General Public License
 # along with ec2publishimg. If not, see <http://www.gnu.org/licenses/>.
 
-import boto3
 import argparse
 import os
-import re
 import sys
 
 import ec2utils.ec2utilsutils as utils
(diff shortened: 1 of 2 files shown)


ec2utils/ec2publishimg/lib/ec2utils/ec2publishimg_utils.py
```diff
@@ -0,0 +1,27 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with ec2publishimg. If not, see <http://www.gnu.org/licenses/>.
+
+import re
+
+from itertools import repeat
+
+
+def validate_accounts(self, share_with):
```
rjschwei: This should be validate_account_numbers, we do not validate the account.

ec2utils/ec2publishimg/ec2publishimg
```diff
@@ -25,6 +25,7 @@ import sys
 
 import ec2utils.ec2utilsutils as utils
 import ec2utils.ec2publishimg as ec2pubimg
+from ec2utils.ec2publishimg_utils import validate_accounts
```
rjschwei: Nope, not doing this. Either validate_account_numbers moves directly into this code or into utilsutils

rjschwei (changes requested):

smarlowucf:
Re-based with changes.

rjschwei (approved):

PR #179, opened 2018-01-10
Implement --create-temporary-security-group option

ChrisBr:
to create a temporary security group with SSH port 22 open.
This is necessary when you do not want to specify a security group but want to
ensure that the helper instance has port 22 open to make it possible to login in via SSH.

This feature is also needed for the OBS cloud uploader. This will make it more convenient for users to upload ec2 images as they don't need to open the port 22 on the default security group or need to create and specify a dedicated security group.

### DIFF
--- a/ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
+++ b/ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
@@ -35,6 +35,7 @@ class EC2ImageUploader(EC2Utils):
                  backing_store='ssd',
                  billing_codes=None,
                  bootkernel=None,
+                 create_temporary_security_group=False,
                  config=None,
                  ena_support=False,
                  image_arch='x86_64',
@@ -64,6 +65,7 @@ class EC2ImageUploader(EC2Utils):
         self.backing_store = backing_store
         self.billing_codes = billing_codes
         self.bootkernel = bootkernel
+        self.create_temporary_security_group = create_temporary_security_group
         self.ena_support = ena_support
         self.image_arch = image_arch
         self.image_description = image_description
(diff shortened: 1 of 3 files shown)


ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
```diff
@@ -238,6 +269,8 @@ def _clean_up(self):
             for volume in self.created_volumes:
                 self._detach_volume(volume, True)
                 self._remove_volume(volume)
+        if self.temporary_security_group_id:
```
ChrisBr: Maybe we need a wait here till the instance is deregisted?
Anyway, I'm not sure if we want to delete the group anyway as for instance the launch-wizard also does not delete the group. @rjschwei what do you think?

ChrisBr: I added now a waiter because otherwise it will fail with a dependency error!

rjschwei:
It makes me think we should approach this from a different direction. Many instance types now only start/run in a VPC, which is why the --subnet-id command line option was added a while back. That of course is useless if the user doesn't have a VPC just yet. Thus logic would dictate that we also add a --create-upload-vpc and then there is another and another and another condition.... Thus it leads to an endless proliferation of commandline options.

If we approach this from the other direction and simply decree that the uploader will setup it's own environment to succeed then all those problems go away. So the uploader can create it's own:
- VPC if non given
- ssh key if non given
- security group

Since we already have command line options for VPC and ssh key I do not want to remove them but I do not want to add anymore command line options that are setup related as it just becomes and endless mess and an already large and convoluted command line becomes unusable.

There should be a basic prepare_aws_for_upload method on the uploader class that creates the resources needed and the existing cleanup method gains the functionality to clean up the created resources.

Can you please proceed along those lines?

ChrisBr:
@rjschwei you're right! We're already working on a temporary key feature because we will need this as well for the upload. We will combine everything!

cc @bgeuken

bgeuken:
The temporary key pair creation is added in #181

rjschwei:
Lets not combine everything at once, since #181 is in process lets get that in and then add methods for
_create_upload_vpc and _create_upload_security group and their counter parts. Each in their own pull request.

Closing this.

PR #180, opened 2018-01-11
Refactor assigning of instance helper

bgeuken:
Just some small thing I stumbled over. Hope it makes sense.

### DIFF
--- a/ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
+++ b/ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
@@ -99,13 +99,13 @@ class EC2ImageUploader(EC2Utils):
         self.storage_volume_size = 2 * self.root_volume_size
 
     # ---------------------------------------------------------------------
-    def _attach_volume(self, instance, volume, device=None):
+    def _attach_volume(self, volume, device=None):
         """Attach the given volume to the given instance"""
         if not device:
             device = self._get_next_disk_id()
         self._connect().attach_volume(
                 VolumeId=volume['VolumeId'],
-                InstanceId=instance['InstanceId'],
+                InstanceId=self.helper_instance['InstanceId'],
                 Device=device)
         if self.verbose:
             print('Wait for volume attachment')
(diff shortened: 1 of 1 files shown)


ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
```diff
@@ -99,13 +99,13 @@ def __init__(self,
         self.storage_volume_size = 2 * self.root_volume_size
 
     # ---------------------------------------------------------------------
-    def _attach_volume(self, instance, volume, device=None):
+    def _attach_volume(self, volume, device=None):
```
ChrisBr, line 102: I'm missing the refactoring where you call ``_attach_volume`` (except the calls in this method of course)

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
```diff
@@ -275,18 +275,14 @@ def _create_image_root_volume(self, source):
             self._check_security_groups_exist()
         if self.running_id:
             # When using an already running instance, simply look up this one
-            helper_instance = self._get_helper_instance()
+            self.helper_instance = self._get_helper_instance()
         else:
-            helper_instance = self._launch_helper_instance()
-        self.helper_instance = helper_instance
+            self.helper_instance = self._launch_helper_instance()
```
rjschwei: In this case I'd like to keep the current logic, i.e.

if self.running_id:
            # When using an already running instance, simply look up this one
            helper_instance = self._get_helper_instance()
        else:
            helper_instance = self._launch_helper_instance()
        self.helper_instance = helper_instance

Just a style thing, but also I think "self.helper_instance = self._get_helper_instance()" can be confusing to the reader as it would be easy for the reader to assume that self._get_helper_instance() just returns self.helper_instance. Therefore the current proposed construct can easily be confused as being self referential.

rjschwei (changes requested):

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
bgeuken: Fixed it now

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
bgeuken: Okay. I've dropped the first commit.

rjschwei (approved):

PR #181, opened 2018-01-12
Create temporary key pair if needed

bgeuken:
This will drop the requirement of providing a ssh key pair to the ec2uploadimg script.

When key_pair_name and ssh_private_key_file parameters are both missing, a temporary key pair get's created. If only one of the two parameters was given, it's assumed that the user wants to use an existing key pair and both parameters get validated and used.

### DIFF
--- a/ec2utils/ec2uploadimg/ec2uploadimg
+++ b/ec2utils/ec2uploadimg/ec2uploadimg
@@ -443,9 +443,6 @@ try:
                                                   region,
                                                   'ssh_key_name',
                                                   '--ssh-key-pair')
-        if not key_pair_name:
-            print('Could not determine key pair name', file=sys.stderr)
-            sys.exit(1)
 
         ssh_private_key_file = args.privateKey
         if not ssh_private_key_file:
@@ -454,24 +451,30 @@ try:
                                                          region,
                                                          'ssh_private_key',
                                                          '--private-key-file')
-        if not ssh_private_key_file:
-            print(
-                'Could not determine the private ssh key file',
-                file=sys.stderr
-            )
+        # If key_pair_name and ssh_private_key_file are missing, a temporary key pair
+        # get's created. Otherwise both parameters are required.
+        if key_pair_name or ssh_private_key_file:
+            if not key_pair_name:
+                print('Could not determine key pair name', file=sys.stderr)
+            if not ssh_private_key_file:
+                print(
+                    'Could not determine the private ssh key file',
+                    file=sys.stderr
+                )
             sys.exit(1)
 
-        ssh_private_key_file = os.path.expanduser(ssh_private_key_file)
+        if ssh_private_key_file:
+            ssh_private_key_file = os.path.expanduser(ssh_private_key_file)
 
-        if not os.path.exists(ssh_private_key_file):
-            print(
-                (
-                     'SSH private key file "%s" does not exist'
-                     % ssh_private_key_file
-                ),
-                file=sys.stderr
-            )
-            sys.exit(1)
+            if not os.path.exists(ssh_private_key_file):
+                print(
+                    (
+                         'SSH private key file "%s" does not exist'
+                         % ssh_private_key_file
+                    ),
+                    file=sys.stderr
+                )
+                sys.exit(1)
 
         ssh_user = args.sshUser
         if not ssh_user:
@@ -517,6 +520,7 @@ try:
                           inst_user_name=ssh_user,
                           launch_ami=ami_id,
                           launch_inst_type=inst_type,
+                          region=region,
                           root_volume_size=root_volume_size,
                           running_id=running_id,
                           secret_key=secret_key,
@@ -533,7 +537,6 @@ try:
                           wait_count=args.waitCount
         )
 
-        uploader.set_region(region)
         if args.snapOnly:
             snapshot = uploader.create_snapshot(args.source)
             print('Created snapshot: ', snapshot['SnapshotId'])
--- a/ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
+++ b/ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
@@ -18,6 +18,7 @@
 import boto3
 import os
 import paramiko
+import string
 import sys
 import threading
 import time
@@ -25,6 +26,7 @@ import time
 
 from ec2utils.ec2utils import EC2Utils
 from ec2utils.ec2UtilsExceptions import EC2UploadImgException
+from tempfile import mkstemp
 
 
 class EC2ImageUploader(EC2Utils):
@@ -44,6 +46,7 @@ class EC2ImageUploader(EC2Utils):
                  inst_user_name=None,
                  launch_ami=None,
                  launch_inst_type='m1.small',
+                 region=None,
                  root_volume_size=10,
                  running_id=None,
                  secret_key=None,
@@ -72,6 +75,7 @@ class EC2ImageUploader(EC2Utils):
         self.inst_user_name = inst_user_name
         self.launch_ami_id = launch_ami
         self.launch_ins_type = launch_inst_type
+        self.region = region
         self.root_volume_size = int(root_volume_size)
         self.running_id = running_id
         self.secret_key = secret_key
@@ -94,9 +98,10 @@ class EC2ImageUploader(EC2Utils):
         self.next_device_id = 0
         self.operation_complete = False
         self.percent_transferred = 0
-        self.region = None
         self.ssh_client = None
         self.storage_volume_size = 2 * self.root_volume_size
+
+        self._prepare_aws_for_upload()
 
     # ---------------------------------------------------------------------
     def _attach_volume(self, volume, device=None):
@@ -238,6 +243,8 @@ class EC2ImageUploader(EC2Utils):
             for volume in self.created_volumes:
                 self._detach_volume(volume, True)
                 self._remove_volume(volume)
+        if self.temporary_key_created:
+            self._remove_upload_key_pair()
 
         self.created_volumes = []
         self.instance_ids = []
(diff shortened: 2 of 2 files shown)


ec2utils/ec2uploadimg/ec2uploadimg
```diff
@@ -534,6 +537,7 @@ try:
         )
 
         uploader.set_region(region)
+        uploader.prepare_aws_for_upload()
```
bgeuken: I was trying to call this method in the initializer, but that method depends on the region. And calling self.set_region from [here](https://github.com/SUSE/Enceladus/blob/master/ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py#L100) doesn't seem to work.

If there is a way to make those to calls in EC2ImageUploader, I would prefer that.

ChrisBr: You could hand over the region and just drop the set_region call, no?
As ec2uploadimg calls super and the region parameter gets set in __init__:
https://github.com/SUSE/Enceladus/blob/master/ec2utils/ec2utilsbase/lib/ec2utils/ec2utils.py#L30

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
```diff
@@ -238,6 +242,8 @@ def _clean_up(self):
             for volume in self.created_volumes:
                 self._detach_volume(volume, True)
                 self._remove_volume(volume)
+        if self.temporary_key_created:
```
ChrisBr, line 246: Does this work? Just wondering because I think you will need a waiter here, otherwise it could fail with an DependencyError. Same as here https://github.com/SUSE/Enceladus/pull/179/files#diff-4e140cc0dbecb24f9927a6ce15557b99R277

bgeuken: The create_key_pair call directly returns a response, which includes the created private key. So I don't think we can have a race condition here.
And for delete_key_pair I assume the same. At least the AWS docu doesn't state that this call is handled asynchronously.

But there isn't any error handling right now. I guess I should add that.

ChrisBr: Yes, for creating I don't see a problem but for deleting. If you attach the key to an instance, I guess you're not allowed to delete this key as long as this instance is not terminated. See LOC 240, the termination usually takes some time (a few minutes). This comment probably makes more sense here https://github.com/SUSE/Enceladus/pull/181/files#diff-4e140cc0dbecb24f9927a6ce15557b99R438

ChrisBr: Maybe just try to delete a key from a running instance, I bet ec2 is complaining about that ... (if not I would be surprised because there is no way to login to this machine anymore...)

ec2utils/ec2uploadimg/ec2uploadimg
bgeuken: That works :+1:

rjschwei: I'm ok with making region part of the constructor arguments and dropping the call.

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
rjschwei: A key is not attached to an instances, what happens is that the public key is transferred into the instance. The key-pair has to exists by name when an instance is created, but it need not live until all instances using the keypair have been terminated. There should not be a problem of deleting the keypair. Anyway, by the time we are cleaning up the helper instance is on it's way out, i.e. it ie being terminated. Thus there's no problem.

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
```diff
@@ -268,6 +277,7 @@ def _create_block_device_map(self, snapshot):
     # ---------------------------------------------------------------------
     def _create_image_root_volume(self, source):
         """Create a root volume from the image"""
+        temporary_key_created = False
```
rjschwei: ?? why is this needed here?

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
```diff
@@ -367,6 +377,22 @@ def _create_target_root_volume(self):
            the image we are creating."""
         return self._create_volume('%s' % self.root_volume_size)
 
+    # ---------------------------------------------------------------------
+    def _create_temporary_private_key(self, key_name='temporary_ec2_uploadkey'):
```
rjschwei: Change method name to _create_upload_key_pair

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
```diff
@@ -367,6 +377,22 @@ def _create_target_root_volume(self):
...
+        if self.verbose:
+            print('Creating temporary key pair')
+        suffix = ''.join(random.choices(string.ascii_uppercase + string.digits, k=12))
+        self.ssh_key_pair_name = key_name + '-' + suffix
+        self.ssh_key_private_key_file = os.path.join('/tmp/', self.ssh_key_pair_name + '.key')
```
rjschwei: We should not write the private key to tmp. We should create a tmpfile and keep it in memory or write it to the users home directory. tmpfile is preferred

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
```diff
@@ -367,6 +377,22 @@ def _create_target_root_volume(self):
...
+        secret_key_content = self._connect().create_key_pair(KeyName=self.ssh_key_pair_name)
+        if self.verbose:
+            print('Write temporary key to ', self.ssh_key_private_key_file)
+        file = open(self.ssh_key_private_key_file, 'w')
+        file.write(secret_key_content['KeyMaterial'])
```
rjschwei: Here the content of the private key will be leaked to every user on the system. Because "open with 'w'" creates the fie with the default mask, which is generally 644, but the private key file needs to be 600. The file needs to be opened with 600 and should be kept in memory.

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
```diff
@@ -408,6 +434,16 @@ def _create_volume(self, size):
 
         return volume
 
+    # ---------------------------------------------------------------------
+    def _destroy_temporary_private_key(self):
```
rjschwei: Rename to _remove_upload_key_pair

rjschwei (changes requested):

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
bgeuken: It's possible, but not recommended for the reasons you just mentioned^^ And given that the helper instance get's terminated, this should be fine.

ChrisBr: Thanks @rjschwei for the clarification!

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
bgeuken: It's a leftover from the previous implementation. I forgot to remove it again.

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
bgeuken: Ouch, you are right :worried:

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
```diff
@@ -364,6 +372,23 @@ def _create_target_root_volume(self):
            the image we are creating."""
         return self._create_volume('%s' % self.root_volume_size)
 
+    # ---------------------------------------------------------------------
+    def _create_upload_key_pair(self, key_name='temporary_ec2_uploadkey'):
+        if self.verbose:
+            print('Creating temporary key pair')
+        home_dir = os.path.expanduser('~/')
+        fd, location = mkstemp(prefix='temporary_ec2_uploadkey.', suffix='.key', dir=home_dir)
```
bgeuken, line 379: @rjschwei I didn't find a way to provide a TemporaryFile object. So it seems we need to create a file on the harddrive. Permissions of that file will be 600 and it's created in the user's home dir.

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
```diff
@@ -53,6 +56,7 @@ def __init__(self,
                  ssh_key_pair_name=None,
                  ssh_key_private_key_file=None,
                  ssh_timeout=300,
+                 temporary_key_created=False,
```
ChrisBr: This is only used internally so I wouldn't expose it to the constructor parameters

bgeuken: Right. Thanks!

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
```diff
@@ -18,7 +18,9 @@
 import boto3
 import os
 import paramiko
+import string
 import sys
+from tempfile import mkstemp
```
rjschwei: Style nitpick, please move "from" form of import after the "import"

rjschwei (changes requested):

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
bgeuken: Okay, done.

rjschwei (approved):

PR #182, opened 2018-01-16
Check XOR key_pair_name and ssh_private_key_file

ChrisBr:
otherwise it will fail silently.

### DIFF
--- a/ec2utils/ec2uploadimg/ec2uploadimg
+++ b/ec2utils/ec2uploadimg/ec2uploadimg
@@ -453,14 +453,11 @@ try:
                                                          '--private-key-file')
         # If key_pair_name and ssh_private_key_file are missing, a temporary key pair
         # get's created. Otherwise both parameters are required.
-        if key_pair_name or ssh_private_key_file:
-            if not key_pair_name:
-                print('Could not determine key pair name', file=sys.stderr)
-            if not ssh_private_key_file:
-                print(
-                    'Could not determine the private ssh key file',
-                    file=sys.stderr
-                )
+        if key_pair_name and not ssh_private_key_file:
+            print('Could not determine the private ssh key file', file=sys.stderr)
+            sys.exit(1)
+        if ssh_private_key_file and not key_pair_name:
+            print('Could not determine key pair name', file=sys.stderr)
             sys.exit(1)
 
         if ssh_private_key_file:


ec2utils/ec2uploadimg/ec2uploadimg
```diff
@@ -453,7 +453,7 @@ try:
                                                          '--private-key-file')
         # If key_pair_name and ssh_private_key_file are missing, a temporary key pair
         # get's created. Otherwise both parameters are required.
-        if key_pair_name or ssh_private_key_file:
+        if (key_pair_name and not ssh_private_key_file) or (not key_pair_name and ssh_private_key_file):
```
bgeuken: If you write it this way you can split the condition and move one part to line 457 and the other to line 459.

But why is the previous one failing? It should have the same effect, (Either key_pair_name and ssh_private_key_file are provided or none of both).

bgeuken: And I would write the second part of the condition as ` ssh_private_key_file and not key_pair_name

ChrisBr: No, what you describe is XOR (Either key_pair_name and ssh_private_key_file are provided or none of both). Or is either key_pair_name OR ssh_private_key_file OR both, you don't want the last case (it will fail silently because both conditions which are nested are false...)

bgeuken: Ah, alright. But than we probably want ` if key_pair_name ^ ssh_private_key_file: ` instead.

Good catch:+1:

ChrisBr: ^ unfortunately does not work with undefined variables (also not for strings, we would need to cast it to boolean ``bool(foo) ^ bool(bar)``), we would need to define it then in the constructor.

bgeuken: That sucks.

Still, I this would be nicer / easier to read:

```python
if ssh_private_key_file and not key_pair_name:
                  print('Could not determine key pair name', file=sys.stderr)

if key_pair_name and not ssh_private_key_file:
                 print(
                     'Could not determine the private ssh key file',
                     file=sys.stderr
                 )

```

ChrisBr: Sure, I can change that!

ChrisBr:
Updated @bgeuken

rjschwei (approved):

PR #183, opened 2018-01-16
Fix attribute definition

ChrisBr:
otherwise it will fail with
'EC2ImageUploader' object has no attribute 'temporary_key_created'


bgeuken:
Looks good. Thanks. And sorry for the breakage.

PR #184, opened 2018-01-17
cloud-netconfig: spec file fixes

jgleissner:
Added dependency on curl
Removed dependency on udev-persistent-ifnames (bsc#1075484)
Use proper self-conflict to make different flavors explicitely conflict with
each other (otherproviders() work-around still needed on SLE11 since its RPM
version does not support self-conflict).

### DIFF
--- a/cloud-netconfig/cloud-netconfig.spec
+++ b/cloud-netconfig/cloud-netconfig.spec
@@ -25,14 +25,10 @@ ExclusiveArch:  do-not-build
 %if "@BUILD_FLAVOR@" == "azure"
 %define flavor_suffix -azure
 %define csp_string Microsoft Azure
-Provides:       cloud-netconfig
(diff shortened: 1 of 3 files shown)


cloud-netconfig/cloud-netconfig.spec
```diff
@@ -55,9 +51,15 @@ Requires:       sysconfig-netconfig
 %endif
 BuildRequires:  udev
 Requires:       udev
-# persistent net generator was split off from udev in Leap
-%if 0%{?leap_version} != 0
-Requires:       udev-persistent-ifnames
+Requires:       curl
+%if 0%{?sles_version} == 11
+# RPM in SLES 11 does not support self conflict, user otherproviders()
```
rjschwei: "use" not "user"

rjschwei (changes requested):

rjschwei (approved):

PR #185, opened 2018-01-17
Check for accountName

ChrisBr:
otherwise fetching from config file will fail.
This is necessary to make creation of temporary SSH key work.

### DIFF
--- a/ec2utils/ec2uploadimg/ec2uploadimg
+++ b/ec2utils/ec2uploadimg/ec2uploadimg
@@ -437,7 +437,7 @@ try:
                                               '--type')
 
         key_pair_name = args.sshName
-        if not key_pair_name:
+        if not key_pair_name and args.accountName:
             key_pair_name = utils.get_from_config(args.accountName,
                                                   config,
                                                   region,
@@ -445,7 +445,7 @@ try:
                                                   '--ssh-key-pair')
 
         ssh_private_key_file = args.privateKey
-        if not ssh_private_key_file:
+        if not ssh_private_key_file and args.accountName:
             ssh_private_key_file = utils.get_from_config(args.accountName,
                                                          config,
                                                          region,


ChrisBr:
Otherwise this will fail:
https://github.com/SUSE/Enceladus/blob/master/ec2utils/ec2utilsbase/lib/ec2utils/ec2utilsutils.py#L163

@rjschwei I'm also thinking about removing the check from ec2utilsutils but I'm not sure which other config options exist here. What do you think?

rjschwei:
@ChrisBr I do not quite understand why this cosmetic change is necessary for your usecase.

https://github.com/SUSE/Enceladus/blob/master/ec2utils/ec2utilsbase/lib/ec2utils/ec2utilsutils.py#L34

checks if we have the necessary information to access the EC2 account, if the information exists then '--account' is not treated as a required argument to the command line.

For the uploader additional checks need to be implemented, i.e. we need to have an ssh-key path and we need to have a key-pair name, i.e. "'--private-key-file" and "--ssh-key-pair" must both be given on the command line or the '--account' option should be treated as required.

The current PR addresses a symptom, but the root cause is a missing check of the basic requirements. That's what should be implemented, this should be in ec2uploadimg command line processor.

The check cannot be removed as it applies to other ec2 command line tools.

rjschwei (changes requested):
See general comment

ChrisBr:
> For the uploader additional checks need to be implemented, i.e. we need to have an ssh-key path and we need to have a key-pair name, i.e. "'--private-key-file" and "--ssh-key-pair" must both be given on the command line or the '--account' option should be treated as required.

That is the problem. In the case that we want to create a temporary key pair, neither a "'--private-key-file" nor a "--ssh-key-pair" is giving which results in --account is required. However, a account should also not be required in this case as we will create a temporary key pair.

I'm not quite sure how you suggest to fix / implement it?

rjschwei:
@ChrisBr right.

Well the original premise of the implementation was that by the time we get to instantiate EC2ImageUploader everything is in order. By me accepting the introduction of _prepare_aws_for_upload() this underlying assumption has been broken. My fault I should have recognized where that was going.

Thus lets go back to the initial assumption and make sure that the AWS account is ready by the time EC2ImageUploader is instantiated. This means _prepare_aws_for_upload() gets removed from the EC2ImageUploader class. ec2utilsutils gets functions to create an ssh key and a vpc. The in the command line processor ec2uploadimg one can call utils.create_ssh_key_pair() etc. Set a flag to clean those up if they have been created in the command line processor. And ec2utilsutils get delete_ssh_key_pair()

ChrisBr:
thanks @rjschwei, will close this one for now and open a new one!

PR #186, opened 2018-01-17
Check for write access before storing key in home directory

ChrisBr:
and otherwise store it in /tmp/

@rjschwei I remember that you suggested to store the key in the home directory the other day. However, I think storing it in /tmp/ with correct permissions should also be fine. 

Background: The build service worker does not have a real home directory. We set the home to ``/etc/obs/`` to store all the configuration files. However, the worker only has read access to this directory so creating a temporary key won't work. Therefore I would like to introduce a fallback directory.

### DIFF
--- a/ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
+++ b/ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
@@ -376,8 +376,10 @@ class EC2ImageUploader(EC2Utils):
     def _create_upload_key_pair(self, key_name='temporary_ec2_uploadkey'):
         if self.verbose:
             print('Creating temporary key pair')
-        home_dir = os.path.expanduser('~/')
-        fd, location = mkstemp(prefix='temporary_ec2_uploadkey.', suffix='.key', dir=home_dir)
+        dir_path = os.path.expanduser('~/')
+        if not os.access(dir_path, os.W_OK):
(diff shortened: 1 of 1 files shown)


ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
```diff
@@ -376,8 +376,10 @@ def _create_target_root_volume(self):
     def _create_upload_key_pair(self, key_name='temporary_ec2_uploadkey'):
         if self.verbose:
             print('Creating temporary key pair')
-        home_dir = os.path.expanduser('~/')
-        fd, location = mkstemp(prefix='temporary_ec2_uploadkey.', suffix='.key', dir=home_dir)
+        dir_path = os.path.expanduser('~/')
+        if not os.access(dir_path, os.W_OK):
+            dir_path = '/tmp/'
```
rjschwei, line 381: I am not OK with /tmp, I understand the need, use mkdtemp()

rjschwei (changes requested):

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
ChrisBr: @rjschwei ok

ChrisBr:
I will combine this one with #185, closing ....

PR #187, opened 2018-01-18
Extract EC2Setup class

ChrisBr:
and move temporary key creation to it.
EC2Preparer#prepare will also create a temporary VPC and Security group with all dependencies.
This makes sure that all dependencies are already set up when calling EC2ImageUploader.

Instead of adding this functionality to the ec2utilsutils class I extracted a new EC2Preparer class to follow the single responsibility principle. Also this class contains a lot of data (like all the temporary assets it creates) which would bloat ec2utilsutils. @rjschwei let me know what you think, I'm open for your suggestions.

### DIFF
--- a/ec2utils/ec2uploadimg/ec2uploadimg
+++ b/ec2utils/ec2uploadimg/ec2uploadimg
@@ -26,6 +26,7 @@ import sys
 import ec2utils.ec2utilsutils as utils
 import ec2utils.ec2uploadimg as ec2upimg
 from ec2utils.ec2UtilsExceptions import *
+from ec2utils.ec2setup import EC2Setup
 
 # Set up command line argument parsing
 argparse = argparse.ArgumentParser(description='Upload image to Amazon EC2')
@@ -359,6 +360,7 @@ if (
     print('--instance-id, or --boot-kernel', file=sys.stderr)
     sys.exit(1)
 
+setup = EC2Setup(access_key, region, secret_key, args.sessionToken, args.verbose)
 try:
     ami_id = args.amiID
     running_id = args.runningID
@@ -437,41 +439,43 @@ try:
                                               '--type')
 
         key_pair_name = args.sshName
+        ssh_private_key_file = args.privateKey
+        if not key_name and not ssh_private_key_file and not args.accountName:
+            key_pair_name, ssh_private_key_file = setup.create_upload_key_pair()
         if not key_pair_name:
             key_pair_name = utils.get_from_config(args.accountName,
                                                   config,
                                                   region,
                                                   'ssh_key_name',
                                                   '--ssh-key-pair')
+        if not key_pair_name:
+            print('Could not determine key pair name', file=sys.stderr)
+            sys.exit(1)
 
-        ssh_private_key_file = args.privateKey
         if not ssh_private_key_file:
             ssh_private_key_file = utils.get_from_config(args.accountName,
                                                          config,
                                                          region,
                                                          'ssh_private_key',
                                                          '--private-key-file')
-        # If key_pair_name and ssh_private_key_file are missing, a temporary key pair
-        # get's created. Otherwise both parameters are required.
-        if key_pair_name and not ssh_private_key_file:
-            print('Could not determine the private ssh key file', file=sys.stderr)
-            sys.exit(1)
-        if ssh_private_key_file and not key_pair_name:
-            print('Could not determine key pair name', file=sys.stderr)
-            sys.exit(1)
 
-        if ssh_private_key_file:
-            ssh_private_key_file = os.path.expanduser(ssh_private_key_file)
+        if not ssh_private_key_file:
+            print(
+                'Could not determine the private ssh key file',
+                file=sys.stderr
+            )
 
-            if not os.path.exists(ssh_private_key_file):
-                print(
-                    (
-                         'SSH private key file "%s" does not exist'
-                         % ssh_private_key_file
-                    ),
-                    file=sys.stderr
-                )
-                sys.exit(1)
+        ssh_private_key_file = os.path.expanduser(ssh_private_key_file)
+
+        if not os.path.exists(ssh_private_key_file):
+            print(
+                (
+                     'SSH private key file "%s" does not exist'
+                     % ssh_private_key_file
+                ),
+                file=sys.stderr
+            )
+            sys.exit(1)
 
         ssh_user = args.sshUser
         if not ssh_user:
@@ -485,6 +489,8 @@ try:
             sys.exit(1)
 
         vpc_subnet_id = args.vpcSubnetId
+        if not vpc_subnet_id and not args.accountName:
+            vpc_subnet_id = setup.create_vpc_subnet()
         if not vpc_subnet_id and not (args.amiID or args.runningID):
             # Depending on instance type an instance may possibly only
             # launch inside a subnet. Look in the config for a subnet if none
@@ -504,6 +510,10 @@ try:
                     msg += '"subnet_id_%s" value' % region
                     print(msg)
 
+        security_group_ids = args.securityGroupIds
+        if not security_group_ids:
+            security_group_ids = setup.create_security_group()
+
         uploader = ec2upimg.EC2ImageUploader(
                           access_key=access_key,
                           backing_store=args.backingStore,
@@ -521,7 +531,7 @@ try:
                           root_volume_size=root_volume_size,
                           running_id=running_id,
                           secret_key=secret_key,
-                          security_group_ids=args.securityGroupIds,
+                          security_group_ids=security_group_ids,
                           session_token=args.sessionToken,
                           sriov_type=sriov_type,
                           ssh_key_pair_name=key_pair_name,
@@ -549,3 +559,6 @@ except EC2UploadImgException as e:
 except Exception as e:
     print(format(e), file=sys.stderr)
     sys.exit(1)
+finally:
+    if setup:
+        setup.clean_up()
--- a/ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
+++ b/ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
@@ -26,7 +26,6 @@ import time
 
 from ec2utils.ec2utils import EC2Utils
 from ec2utils.ec2UtilsExceptions import EC2UploadImgException
-from tempfile import mkstemp
 
 
 class EC2ImageUploader(EC2Utils):
@@ -100,9 +99,6 @@ class EC2ImageUploader(EC2Utils):
         self.percent_transferred = 0
         self.ssh_client = None
         self.storage_volume_size = 2 * self.root_volume_size
-        self.temporary_key_created = False
-
-        self._prepare_aws_for_upload()
 
     # ---------------------------------------------------------------------
     def _attach_volume(self, volume, device=None):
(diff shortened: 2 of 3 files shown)


ec2utils/ec2uploadimg/ec2uploadimg
```diff
@@ -26,6 +26,7 @@ import sys
 import ec2utils.ec2utilsutils as utils
 import ec2utils.ec2uploadimg as ec2upimg
 from ec2utils.ec2UtilsExceptions import *
+from ec2utils.ec2preparer import EC2Preparer
```
rjschwei: Lets call this ec2setup and correspondingly EC2Setup

ec2utils/ec2uploadimg/ec2uploadimg
```diff
@@ -436,42 +438,44 @@ try:
                                               'instance_type',
                                               '--type')
 
-        key_pair_name = args.sshName
+        preparer = EC2Preparer(access_key, args.accountName, args.sshName, region, secret_key, args.securityGroupIds, args.sessionToken, args.privateKey, args.vpcSubnetId, True).prepare()
```
rjschwei: The Setup class should be constructed just with the access_key, secrete_key, and region. The when we determine something is missing i.e. 

```
if not key_pair_name and not args.accountName:
     key_pair_name, ssh_private_key_file = perparer.create_ssh_key_pair()
```

Then the other logic in the code pretty much can stay the same. Thus our decision tree ends up like this:
- if keyname given use it
- if no keyname given and no account -> generate
- if no keyname given but account -> get from config
- else fail

ec2utils/ec2uploadimg/ec2uploadimg
```diff
@@ -484,7 +488,7 @@ try:
             print('Could not determin ssh user to use', file=sys.stderr)
             sys.exit(1)
 
-        vpc_subnet_id = args.vpcSubnetId
+        vpc_subnet_id = preparer.vpc_subnet_id
```
rjschwei: Similar to the above there'd be a check and then a call 

vpc_subnet_id = preparer.create_vpc()

Having the individual checks for the things that might be needed allows us to avoid creating unnecessary pieces, for example if a user wants to use a temporary key but has a specific subnet for the upload. The positive side effect is also that the constructor for the Setup class has fewer arguments.

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
```diff
@@ -240,6 +239,29 @@ def _clean_up(self):
...
+                    )
+                except:
+                    wait_status = 1
+                if self.verbose:
+                    self.progress_timer.cancel()
+                repeat_count = self._check_wait_status(
+                    wait_status,
+                    error_msg,
+                    repeat_count
+                )
```
rjschwei, line 261: This is correct, just not certain if we really need/want to wait for the instance to terminate. Volume detach and delete works even if the platform considers the instance running. The only thing this would do is "protect" the user from incurring costs if an instance fails to properly shut down. But I have not ever seen this. Well I have seen instances take a longtime to terminate, but they eventually all do. Thus I am not certain we need to wait here.

Below the check for "temporary_key_created should be removed as this is now in the EC2Setup class implementation.

ec2utils/ec2utilsbase/lib/ec2utils/ec2preparer.py
```diff
@@ -0,0 +1,173 @@
...
+import boto3
+import os
+import random
+import datetime
+
+from ec2utils.ec2UtilsExceptions import EC2ConnectionException
+from ec2utils.ec2utils import EC2Utils
+from tempfile import mkstemp
+
+class EC2Preparer(EC2Utils):
```
rjschwei: EC2Setup ;)

ec2utils/ec2utilsbase/lib/ec2utils/ec2preparer.py
```diff
@@ -0,0 +1,173 @@
...
+    """Class to prepare an Amazon EC2 account with all necessary resources"""
+
+    def __init__(self, access_key, account, key_pair_name, region, secret_key, security_group_ids, session_token, ssh_private_key_file, vpc_subnet_id, verbose):
```
rjschwei: Simplify the constructor, see previous comment

Derive from EC2UtilsBase then you can take advantage of creating the connection to EC2 that is already implemented there.

ec2utils/ec2utilsbase/lib/ec2utils/ec2preparer.py
```diff
@@ -0,0 +1,173 @@
...
+    def clean_up(self):
+        if self.temporary_key_created:
+            self._remove_upload_key_pair()
+        if self.temporary_security_group_created:
+            self._remove_security_group()
+        if self.temporary_vpc_created:
+            self._remove_vpc()
+
+    # ---------------------------------------------------------------------
+    def _create_upload_key_pair(self, key_name='temporary_ec2_uploadkey'):
```
rjschwei: Becomes a public method and should create it's own connection using ec2utils base class

ec2utils/ec2utilsbase/lib/ec2utils/ec2preparer.py
```diff
@@ -0,0 +1,173 @@
...
+                 'FromPort': 22,
+                 'ToPort': 22,
+                 'IpRanges': [{'CidrIp': '0.0.0.0/0'}]}
+            ])
+
+        if self.verbose:
+            print("Successfully allowed incoming SSH port 22 for security group %s in %s" % (self.security_group_ids, self.vpc_id))
+
+    # ---------------------------------------------------------------------
+    def _create_vpc(self):
```
rjschwei: Public

ec2utils/ec2utilsbase/lib/ec2utils/ec2preparer.py
```diff
@@ -0,0 +1,173 @@
...
+    def _create_vpc_subnet(self):
+        response = self._connect().create_subnet(CidrBlock='192.168.1.0/24', VpcId=self.vpc_id)
+        self.vpc_subnet_id = response['Subnet']['SubnetId']
+        self._connect().associate_route_table(SubnetId=self.vpc_subnet_id, RouteTableId=self.route_table_id)
+        self._connect().modify_subnet_attribute(MapPublicIpOnLaunch={ 'Value': True }, SubnetId=self.vpc_subnet_id)
+        if self.verbose:
+            print("Created VPC subnet with id %s" % (self.vpc_subnet_id))
+
+    # ---------------------------------------------------------------------
+    def prepare(self):
```
rjschwei: drop this logic and incorporate it into ec2uploadimg command line processor. See previous comments

rjschwei (changes requested):

ec2utils/ec2uploadimg/ec2uploadimg
ChrisBr: That already happened before because the check about the pieces happened in the Ec2Preparer class but I'm fine with changing it.

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
ChrisBr: This is necessary because of deleting the temporary VPC and Security Group. If you want to delete the VPC / Security Group of a still running (or stopped) instance, you will get a DependentResource error. However, we could only do that in case there is a temporary resource but then either

a) we need to hand over to EC2ImageUploader if there is a temporary resource (don't like that)
or
b) do it in the EC2Setup class. But then we need to hand over the helper instance(s) ID(s) to the EC2Setup class (don't like that either)
So in either way, we need to handover information from one class to another. So I thought simplest solution is to always wait till the instance is properly terminated.

ec2utils/ec2utilsbase/lib/ec2utils/ec2preparer.py
ChrisBr: That's what I already do, no? I inherit from EC2Utils and use the _connect method from it? Or what do you mean exactly?

ec2utils/ec2utilsbase/lib/ec2utils/ec2preparer.py
ChrisBr: I'm already using a dedicated connection from ec2utils by using the _connect method, no?

ec2utils/ec2uploadimg/ec2uploadimg
```diff
@@ -437,41 +439,43 @@ try:
...
-        if key_pair_name and not ssh_private_key_file:
-            print('Could not determine the private ssh key file', file=sys.stderr)
-            sys.exit(1)
-        if ssh_private_key_file and not key_pair_name:
-            print('Could not determine key pair name', file=sys.stderr)
-            sys.exit(1)
 
-        if ssh_private_key_file:
-            ssh_private_key_file = os.path.expanduser(ssh_private_key_file)
+        if not ssh_private_key_file:
```
ChrisBr, line 462: Note that this is basically just a revert of https://github.com/SUSE/Enceladus/pull/181

ChrisBr:
@rjschwei I tried to follow all your suggestions, please have a look again. Please also not my comments inline.

Please note that I haven't tried the changes you suggested yet. So please ping me when you think this is ready to merge but do not merge so that I can quickly try it out again.

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
rjschwei: agreed

ec2utils/ec2uploadimg/ec2uploadimg
```diff
@@ -549,3 +559,6 @@ except EC2UploadImgException as e:
 except Exception as e:
     print(format(e), file=sys.stderr)
     sys.exit(1)
+finally:
+    if setup:
```
rjschwei, line 563: Should call clean_up unconditionally as there is no condition around the creation of the class and we do not know if any temporary resources were created. Other option would be

if not args.accountName:
  setup = ....

earlier in the code, but that would imply that we have to carry if tests in other places. Thus I prefer unconditional creation of the EC@Setup() instance and then unconditional call of clean_up().

ec2utils/ec2uploadimg/lib/ec2utils/ec2uploadimg.py
rjschwei: agreed

ec2utils/ec2utilsbase/lib/ec2utils/ec2setup.py
```diff
@@ -0,0 +1,183 @@
...
+            GroupId=self.security_group_id,
+            IpPermissions=[
+                {'IpProtocol': 'tcp',
+                 'FromPort': 22,
+                 'ToPort': 22,
+                 'IpRanges': [{'CidrIp': '0.0.0.0/0'}]}
+            ])
+
+        if self.verbose:
+            print("Successfully allowed incoming SSH port 22 for security group %s in %s" % (self.security_group_id, self.vpc_id))
```
rjschwei, line 80: Line length ;) pep8 will complain.

ec2utils/ec2utilsbase/lib/ec2utils/ec2setup.py
```diff
@@ -0,0 +1,183 @@
...
+        self.internet_gateway_id = response['InternetGateway']['InternetGatewayId']
+        self._connect().attach_internet_gateway(VpcId=self.vpc_id, InternetGatewayId=self.internet_gateway_id)
+        if self.verbose:
+            print("Successfully created internet gateway %s" % (self.internet_gateway_id))
+
+    # ---------------------------------------------------------------------
+    def _create_route_table(self):
+        response = self._connect().create_route_table(VpcId=self.vpc_id)
+        self.route_table_id = response['RouteTable']['RouteTableId']
+        self._connect().create_route(DestinationCidrBlock='0.0.0.0/0', GatewayId=self.internet_gateway_id, RouteTableId=self.route_table_id)
```
rjschwei, line 124: line length ;)

rjschwei (changes requested):
Getting close, thanks. Please run your code through pep8 after the changes.

ec2utils/ec2uploadimg/ec2uploadimg
ChrisBr: Removing the conditional should be fine, right! This is a left over from a previous iteration as we created the EC2Setup object later. Now we create it before the the try / except block so it should always be there.

mschnitzer:
@rjschwei I also found some syntax errors which are already fixed but not updated in this PR. I'll create a new one tomorrow with your feedback you just have provided. Thanks!

(@ChrisBr as info for you)

ChrisBr:
superseded by https://github.com/SUSE/Enceladus/pull/188

ec2utils/ec2uploadimg/ec2uploadimg
ChrisBr: The conditional is actually necessary as the setup is created inside the regions for loop which means if there is no region the setup is not defined and would cause a null pointer exception.

PR #188, opened 2018-01-24
Extract EC2Setup class (new)

mschnitzer:
and move temporary key creation to it.
EC2Preparer#prepare will also create a temporary VPC and Security group with all dependencies.
This makes sure that all dependencies are already set up when calling EC2ImageUploader.

Instead of adding this functionality to the ec2utilsutils class I extracted a new EC2Preparer class to follow the single responsibility principle. Also this class contains a lot of data (like all the temporary assets it creates) which would bloat ec2utilsutils. @rjschwei let me know what you think, I'm open for your suggestions.

_________

@rjschwei I addressed all pep8 issues and fixed a syntax error. Please have a look again. I will squash the commits before the merge. It's just for you to see what has changed.

Both files were put through `pep8`.

_________

Closes #187

### DIFF
--- a/ec2utils/ec2uploadimg/ec2uploadimg
+++ b/ec2utils/ec2uploadimg/ec2uploadimg
@@ -360,213 +360,211 @@ if (
     print('--instance-id, or --boot-kernel', file=sys.stderr)
     sys.exit(1)
 
-setup = None
-
 try:
     ami_id = args.amiID
     running_id = args.runningID
     for region in regions:
-        setup = EC2Setup(access_key, region, secret_key, args.sessionToken,
-                         args.verbose)
-        if not ami_id and not running_id:
-            try:
-                ami_id = utils.get_from_config(args.accountName,
-                                               config,
-                                               region,
-                                               'ami',
-                                               '--ec2-ami')
-            except:
-                print('Could not determine helper AMI-ID', file=sys.stderr)
-                sys.exit(1)
-        bootkernel = args.akiID
-        if args.virtType == 'hvm':
-            bootkernel = None
-
-        if not bootkernel and args.virtType != 'hvm':
-            if args.grub2:
+        try:
+            setup = EC2Setup(access_key, region, secret_key, args.sessionToken,
+                             args.verbose)
+            if not ami_id and not running_id:
                 try:
-                    bootkernel = utils.get_from_config(args.accountName,
-                                                       config,
-                                                       region,
-                                                       'g2_aki_x86_64',
-                                                       '--boot-kernel')
+                    ami_id = utils.get_from_config(args.accountName,
+                                                   config,
+                                                   region,
+                                                   'ami',
+                                                   '--ec2-ami')
                 except:
+                    print('Could not determine helper AMI-ID', file=sys.stderr)
+                    sys.exit(1)
+            bootkernel = args.akiID
+            if args.virtType == 'hvm':
+                bootkernel = None
+
+            if not bootkernel and args.virtType != 'hvm':
+                if args.grub2:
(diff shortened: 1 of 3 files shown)


mschnitzer:
@rjschwei the code has been tested and it seems to work.

ec2utils/ec2uploadimg/ec2uploadimg
```diff
@@ -437,41 +442,45 @@ try:
                                               '--type')
 
         key_pair_name = args.sshName
+        ssh_private_key_file = args.privateKey
+        if not key_pair_name and not ssh_private_key_file and not \
+                args.accountName:
```
rjschwei: Wrap this condition in (), to match our style, i.e.

if (
    not key_pair_name 
    and not ssh_private_key_file
    and not args.accountName
):

ec2utils/ec2uploadimg/ec2uploadimg
```diff
@@ -437,41 +442,45 @@ try:
...
+            key_pair_name, ssh_private_key_file = \
+                setup.create_upload_key_pair()
```
rjschwei: Should fit on the line above

ec2utils/ec2uploadimg/ec2uploadimg
```diff
@@ -549,3 +564,6 @@ except EC2UploadImgException as e:
 except Exception as e:
     print(format(e), file=sys.stderr)
     sys.exit(1)
+finally:
+    if setup:
+        setup.clean_up()
```
rjschwei: clean up should be unconditional, we create the object unconditionally. Also cleanup needs to be in the region loop or we will leave traces behind in each region but the last. clean_up call also needs to be in the exception handlers.

rjschwei (changes requested):

ec2utils/ec2uploadimg/ec2uploadimg
mschnitzer: Nope as it's `80` characters long. The default line length limit is `79` characters.

ec2utils/ec2uploadimg/ec2uploadimg
```diff
@@ -366,207 +366,208 @@ try:
...
-                print('Could not determine helper AMI-ID', file=sys.stderr)
-                sys.exit(1)
-        bootkernel = args.akiID
-        if args.virtType == 'hvm':
-            bootkernel = None
-
-        if not bootkernel and args.virtType != 'hvm':
-            if args.grub2:
+        setup = None
+        try:
```
ChrisBr: I think the indentation is now wrong!
Also, what is the purpose of this change? I don't get it?
That you can define setup inside the loop?

ec2utils/ec2uploadimg/ec2uploadimg
```diff
@@ -366,207 +366,208 @@ try:
...
-            except:
-                print('Could not determine helper AMI-ID', file=sys.stderr)
-                sys.exit(1)
-        bootkernel = args.akiID
-        if args.virtType == 'hvm':
-            bootkernel = None
-
-        if not bootkernel and args.virtType != 'hvm':
-            if args.grub2:
+        setup = None
```
ChrisBr: This shouldn't be necessary

ec2utils/ec2uploadimg/ec2uploadimg
ChrisBr: Discussed offline, fine for me

mschnitzer:
@rjschwei open for review again ;-)

rjschwei (approved):

PR #190, opened 2018-02-01
Fix json output.

### LINKED ISSUE #189: Pint decode error with --json
Problem: Json output is already formatted to str and throws a decode error.

https://github.com/SUSE/Enceladus/blob/master/susePublicCloudInfoClient/bin/pint#L100 and 107.

Possibly should decode xml directly: https://github.com/SUSE/Enceladus/blob/master/susePublicCloudInfoClient/lib/susepubliccloudinfoclient/infoserverrequests.py#L267

smarlowucf:
Json is already str type. Only decode xml output.

Resolves #189


rjschwei:
Oh well

smarlowucf:
> Oh well

😕

PR #192, opened 2018-02-07
- regionService

rjschwei:
+ Reduce memory foot print. Rather than building a map of all IP addresses
    use the ranges. When a request is received without region hint find
    ranges with longest possible prefix length then verify that the requester
    IP is one of the potential ranges.

### DIFF
--- a/regionService/srv/www/regionService/regionInfo.py
+++ b/regionService/srv/www/regionService/regionInfo.py
@@ -45,13 +45,11 @@ smt-fingerprint = SMT_CERT_FINGERPRINT
 
 import configparser
 import getopt
-import hashlib
 import ipaddress
 import logging
 import os
 import random
 import sys
-import threading
 
 from flask import Flask
 from flask import request
@@ -61,15 +59,15 @@ from flask import request
 def create_smt_region_map(conf):
     """Create two mappings:
          ip_to_smt_data_map:
-             maps all IP addresses to their respctive SMT server info
+             maps all IP ranges to their respctive SMT server info
          region_name_to_smt_data_map:
              maps all region names to their respective SMT server info"""
-    ip_to_smt_data_map = {}
+    ip_range_to_smt_data_map = {}
     region_name_to_smt_data_map = {}
     region_data_cfg = configparser.RawConfigParser()
     try:
         parsed = region_data_cfg.read(conf)
-    except:
+    except Exception:
         logging.error('Could not parse configuration file %s' % conf)
         type, value, tb = sys.exc_info()
         logging.error(value.message)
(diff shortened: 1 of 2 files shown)


regionService/srv/www/regionService/regionInfo.py
```diff
@@ -132,17 +130,43 @@ def create_smt_region_map(conf):
...
+def find_longest_prefix_ipv4(requester, ip_range_map):
+    """Find IP ranges for IPv4 adresses that are likely to contain the
+       requester address"""
+    potential_ranges = {
+        1: [],
+        2: [],
+        3: []
+    }
+    results_map = {}
+    requester_prts = requester.split('.')
```
bear454: How about `requester_parts` ?

regionService/srv/www/regionService/regionInfo.py
```diff
@@ -132,17 +130,43 @@ def create_smt_region_map(conf):
...
+    for ip_range in ip_range_map.keys():
+        ip_range_prts = ip_range.split('.')
```
bear454: Same here - why abbreviate "parts" ?

regionService/srv/www/regionService/regionInfo.py
```diff
@@ -132,17 +130,43 @@ def create_smt_region_map(conf):
...
+        match_cnt = 0
+        for i in range(3):
+            if requester_prts[i] == ip_range_prts[i]:
+                match_cnt += 1
+        if match_cnt:
+            potential_ranges[match_cnt].append(ip_range)
+    for match_len in [3, 2, 1]:
```
smarlowucf: It's not any shorter but could use generator instead of list.

```range(3, 0, -1)```

**Comment by smarlowucf** [comment]
@rjschwei Not sure I follow exactly what we're doing here but...

It looks like we have an IP_Adr and a list of IP_Nets and we need to find the network the address is in? Or is this oversimplified? If not then I wonder what the performance of using a simple `in` clause would be? If it uses a generator it might be a performant option.

```
if ip_address in ip_network:
    # do stuff
```

Edit: regarding ```find_longest_prefix_ipv4```

regionService/srv/www/regionService/regionInfo.py
rjschwei: So my finger skin doesn't wear off, fixing this, it's habbitt

rjschwei:
Here are the performance numbers:
Execution time no prefix search:  2.5799427032470703
Execution time with prefix search:  0.24348831176757812

Where "no prefix search" means:

`if ip_address in ip_network:`

The test conditions are as follows:
- generate a list of requesters
  + the requesters are 1 IP address for every range of the current AWS data set of IP ranges
- loop over all requesters and find the smt data
- no action is taken when the IP range is found that contains the requester

Searching for the matching prefix first is significantly faster.

For IPv6 support this will change again, thanks to @bear454 finding https://github.com/jsommers/pytricia

Issue #194, opened 2018-02-21
Clean up on SIGTERM


ChrisBr:
As you know, we use ec2uploadimg as the "backend" for the OBS cloud uploader. We also provide the user the possibility to "abort" the upload which currently sends SIGTERM to ec2uploadimg which potentially leaves resources (e.g. helper instance, volumes etc) in AWS left.

I'm opening this issue to discuss some potential solutions to properly clean up when aborting the upload. 

One solution I'm currently trying is to install a [signal handler](https://docs.python.org/2/library/signal.html) in ec2uploadimg. 

```
def signal_handler(signal, frame):
    if uploader:
      uploader.clean_up()
    if setup:
      setup.clean_up()

signal.signal(signal.SIGTERM, signal_handler)
```

However, I don't really like this solution too much and also it seems it is not that reliable.

@rjschwei is a signal hanlder sth you would consider to include into ec2uploadimg? Do you have any other / better solution we could try out?

Thank you

rjschwei:
@ChrisBr I am not opposed to a signal handler but it does have it's limitations. When invoked while we are in the middle of a transfer things are going to get really ugly because we are probably going to be in a place where it is difficult to clean up the sftp connection. Thus in order to do this cleanly with a signal handler we'd have to keep track of the stage of the image creation process and that will be a lot of overhead. Some examples:

- User aborts before upload starts:
  -> Easy, just kill the helper instance and clean up any created resources
- User aborts during image upload
  -> Not so easy, we should shut down the connection properly rather than just killing the receiving end and then cleaning up the ressources
- User aborts after snapshot is created but before image is registered
  -> Now what? Technically we'd have to remove the created snapshot.

I think it is cleaner to let the process run to the end and then clean up. Meaning we still have the signal handler but it just sets a flag on the uploader. The uploader checks the flag as it goes through its stages and then does "the right" thing. 

```
def signal_handler(signal, frame):
    uploader.abort()

signal.signal(signal.SIGTERM, signal_handler)
```
In the uploader itself:

```
def abort(self):
    """Set the abort flag to take appropriate action and stop image creation"""
    self.abort = True
```

Then in the "right" places

```
if self.abort:
......
```

ChrisBr:
@rjschwei thanks a lot for your update. You're right, it is best to just set a flag and clean up in stages.

This is the approach I implemented for now:

- If the upload already started, we finish the current step and run the clean up afterwards.
[...]

PR #196, opened 2018-02-22
Implement AWS clean up when aborting an upload

ChrisBr:
When an upload gets aborted with Ctrl + C or SIGTERM / SIGINT
ec2uploadimg left resources like helper instances left in AWS.
This commit implements proper clean up of these AWS resources.

- If the upload already started, we finish the current step and run the clean up afterwards.
- If the upload did not start but the setup process, we finish the setup and do not start the upload. Finally we clean up the setup (VPC etc).
- If neither setup nor upload started, we will just exit.

If the script gets aborted a second time, we just exit without a clean up like we did before.

Implements #194

### DIFF
--- a/ec2utils/ec2uploadimg/ec2uploadimg
+++ b/ec2utils/ec2uploadimg/ec2uploadimg
@@ -22,11 +22,32 @@ import argparse
 import boto3
 import os
 import sys
+import signal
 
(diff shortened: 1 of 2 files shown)


rjschwei (approved):
LGTM

PR #197, opened 2018-02-26
Add new regions to example config file

ChrisBr:
This adds the new regions

- ca-central-1
- eu-west-2
- eu-west-3
- ap-south-1

@rjschwei should we add aki_i386, aki_x86_64 = aki-880531cd and g2_aki_x86_64 as well?

### DIFF
--- a/ec2utils/ec2utils.conf.example
+++ b/ec2utils/ec2utils.conf.example
@@ -164,3 +164,43 @@ user = ec2-user
 # Allow a region to overwrite the account key
 # ssk_key_name =
 # ssh_private_key =
+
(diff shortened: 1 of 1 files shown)


rjschwei:
While you are at it, please add 

```
[region-cn-northwest-1]
ami = ami-23978241
instance_type = t2.micro
user = ec2-user
# Allow a region to overwrite the account key
# ssk_key_name =
# ssh_private_key =
```

Thanks

rjschwei:
Thanks, sorry for the long delay still struggling to catch up after vacation which is already more than a month in the past.

PR #198, opened 2018-03-21
- regionServiceClient

rjschwei:
+ Update Azure plugin to get the location from metadata server, replacing
    proforma implementation with prrroperrr URL access location


jgleissner (approved):
Looks good to me.

bear454 (approved):
:+1:

PR #201, opened 2018-04-20
- serviceAccessConfig

rjschwei:
+ Properly handle a <RequireAll> directive in the apache config


serviceAccessConfig/lib/serviceAccessConfig/apache.py
```diff
@@ -81,13 +87,25 @@ def _update_service_config(self, cidr_blocks):
...
+        opening_directive = directive_settings[0]
+        ip_directive = directive_settings[1]
+        closing_directive = directive_settings[2]
+        if opening_directive:
+            allow += opening_directive + '\n'
+            indent_mul += 1
+        indentation = space * indent * indent_mul
         while len(cidr_blocks) > 300:
-            allow += '        %s %s\n\n' % (ip_directive,
+            allow += indentation + '%s %s\n\n' % (ip_directive,
```
smarlowucf: indentation could be part of the string `'%s%s %s\n\n'`. And the indent below doesn't match opening bracket `space.join(cidr_blocks[:300]))`.

serviceAccessConfig/lib/serviceAccessConfig/apache.py
```diff
@@ -45,7 +45,9 @@ def _get_apache_ip_directive(self):
         """Figure out the version of Apache and return the appropriate
            Apache directive for IP address filtering"""
         apachectl = None
+        closing_directive = None
```
bear454: Consider using an empty string, instead of `None` to avoid type casting issues (e.g. `TypeError: cannot concatenate 'str' and 'NoneType' objects`)

serviceAccessConfig/lib/serviceAccessConfig/apache.py
```diff
@@ -57,9 +59,11 @@ def _get_apache_ip_directive(self):
             for scr in upgradeScript:
                 toVer = scr.split('-')[2]
                 if int(toVer) >= 24:
+                    opening_directive = '    <RequireAny>'
                     directive = 'Require ip'
+                    closing_directive = '    </RequireAny>'
                     break
-            return directive
+            return opening_directive, directive, closing_directive
```
bear454, line 66: As a matter of style, I don't like multiple unique returns, but given the scope, I don't expect you to change it in this PR.

serviceAccessConfig/lib/serviceAccessConfig/apache.py
smarlowucf: String type default to empty string ''.

rjschwei:
@bear454 @smarlowucf thanks for the comments, realized needed another change to avoid an endless list of opening and closing directives. Another look is appreciated.

serviceAccessConfig/lib/serviceAccessConfig/apache.py
```diff
@@ -126,6 +126,16 @@ def _update_service_config(self, cidr_blocks):
                         found_order = 1
                         if ip_directive == 'Allow from':
                             new_content += order
+                    # Strip opening and closing directive from new
+                    # access restriction list if they exist and are in the
+                    # config file
+                    elif ( ln.startswith(opening_directive) ):
```
bear454: Just to be complete, consider checking that it ends with `closing_directive` as well?

rjschwei: Well closing_directive would be on another line which implies I'd have to set some flag, to result in

if openning_directive_found and ln.strip().startswith(closing_directive):
    allow = .....

I don't think this is necessary because if the closing directive is not there in the first place then Apache would not run and complain a syntax error in the config.

smarlowucf (approved):

PR #206, opened 2018-06-19
- gcemetadata

rjschwei:
+ Support multiple nics, backport of #a6b7cb9


gcemetadata/lib/gcemetadata/gcemetadata.py
```diff
@@ -35,6 +35,7 @@ def __init__(self, apiv='v1'):
         self.default_disk_id = '0'
         self.default_license_id = '0'
         self.default_net_dev_id = '0'
+        self.default_subnet_id = '0'
```
bear454, line 38: What is this used for?

rjschwei: I was following the pattern for net_dev_id. Currently this is not used, however, I have not investigated additional separation on the networking side in GCE, meaning if it is possible to create another subnet on what is essentially nic-1 and then have additional entries in the metadata.

So for consistency I kept this in the backport, same question really applies to original fix. I have opened #207 to make sure we do not loose track of this.

PR #210, opened 2018-06-28
Mv gcemet

rjschwei:


bear454 (approved):
I verified this is now in https://github.com/SUSE-Enceladus/gcemetadata

PR #219, opened 2018-10-12
[WIP] Write IP ranges to /etc/allowed-ips.conf

ikapelyukhin:


serviceAccessConfig/lib/serviceAccessConfig/apache.py
```diff
@@ -107,73 +107,10 @@ def _update_service_config(self, cidr_blocks):
...
-                        # Add the closing tag for the directive
-                        new_content += ln
-                    else:
-                        new_content += ln
-                # Write everything else
-                else:
-                    new_content += ln
-
         # Write the updated config
+        cfg = '/etc/apache2/allowed-ips.conf'
```
ikapelyukhin, line 111: Not sure how easy it is to update `srvAccess.cfg` on every server. `nu_server.conf` would have to be replaced with the stock config from the RPM, too.

ikapelyukhin:
Do we still care about upgrading from Apache 2.2?

rjschwei:
@ikapelyukhin Sorry I moved the code out from under you, it is now in SUSE-Enceladus/service-access-config. Will add the nginx code there, can you bring this along when you have a chance? Thanks

ikapelyukhin:
@rjschwei well, let's just close this PR. We'll move to RMT soon and use nginx code.

PR #222, opened 2018-11-07
- regionServiceClient

rjschwei:
+ Fix dependencies for package
  + Handle exception to avoid traceback
    - The exec_subcommand function either returns a tuple or an int. An
      exception if triggered when a tuple is expected and and int is returned
      when used with the Python automatic unpack feature.


jgleissner (commented):
It'll do, but IMHO I'd not even catch the exception in exec_subprocess at all. The caller, if interested in the process output, is potentially forced into a TypeError exception and hence has to catch anyway, and a caller that is only interested in the exit code could distinguish between an execution failure and a genuine exit code of -1.

rjschwei:
The exception propagated to the top as indicated in the bug is not an exception from executing subprocess, it is triggered by the different return types of the exec_subprocess() convenience function.

So the choices are:
- make the return type of the exec_subprocess() function consistent, i.e. even for a call that only wants the return code return a tuple, such as (-1, -1) or (0,0) for success. That makes everything else more difficult as a simply if test will no longer work; the tuples (-1,-1) and (0, 0) both evaluate to true. Or return either the error string or the return from STDOUT depending on the condition and thus the return is consistent in that it cannot be unpacked.
- Make the client (point of call of exec_subprocess) handle the exception that subprocess potentially throws. In this case having the convenience function of exec_subprocess() makes little sense as we'd need to wrap every call in try: except: and if we do that we might as well just call subprocess directly as it only adds 1 extra line at each call location.

I opted for the in between solution to wrap only the call that is expecting a tuple as a return value in try-except.

jgleissner:
Well, for the return_output==True case you already have the except clause anyway. For the return_output==False case, I guess you could save a couple of lines, **if** you're not interested in doing anything special in case the command couldn't be executed (which I think you normally would).

Anyhow, it'll work fine and it's not like I feel very strongly about it. :)

jgleissner (approved):

PR #224, opened 2019-01-04
- regionServiceClient

rjschwei:
+ Complete IPv6 handling in the registration client. The current
    implementation had various shortcomings and bugs.
    ~ Did not prefer IPv6 over IPv4 when available -> not RFC compliant
    ~ Tried to access undefined attributes when old format SMT was loaded from
      cache generating a traceback
    ~ Messages did not properly reflect IPv6 or IPv4 access path
    ~ hosts file update did not consider IPv6 entries and replacement
      thereof


regionServiceClient/lib/cloudregister/registerutils.py
```diff
@@ -45,8 +45,11 @@ def add_hosts_entry(smt_server):
     smt_hosts_entry_comment += 'retain comment as well\n'
     hosts = open('/etc/hosts', 'a')
     hosts.write(smt_hosts_entry_comment)
+    smt_ip = smt_server.get_ipv4()
```
smarlowucf, line 48: Might as well put this in an else clause. So it's not called when has_ipv6_access is True.

regionServiceClient/lib/cloudregister/registerutils.py
```diff
@@ -378,6 +389,25 @@ def get_zypper_pid():
     return pidData[0].strip().decode()
 
 
+# ----------------------------------------------------------------------------
+def has_ipv6_access(smt):
+    """IPv6 access is possible if we have an SMT server that has an IPv6
+       address and it can be accessed over IPv6"""
+    if not smt.get_ipv6():
+        return None
```
smarlowucf: Could use consistent return types (False).

regionServiceClient/lib/cloudregister/registerutils.py
```diff
@@ -311,11 +318,15 @@ def get_current_smt():
         return
     # Verify that this system is also in /etc/hosts and we are in
     # a consistent state
-    smt_ip = smt.get_ip()
+    smt_ipv4 = smt.get_ipv4()
+    smt_ipv6 = smt.get_ipv6()
```
smarlowucf, line 322: Should this and the re.search be conditional? From below it looks like get_ipv6() could yield None which would mean searching for 'None\s'.

regionServiceClient/lib/cloudregister/registerutils.py
```diff
@@ -527,8 +557,8 @@ def remove_registration_data():
     smt_data_file = __get_registered_smt_file_path()
     if os.path.exists(smt_data_file):
         smt = get_smt_from_store(smt_data_file)
-        ip_address = smt.get_ip()
-        logging.info('Clean current registration server: %s' % ip_address)
+        smt_ips = (smt.get_ipv4(),smt.get_ipv6())
```
smarlowucf: Space after comma for pep8.

regionServiceClient/lib/cloudregister/registerutils.py
```diff
@@ -547,12 +577,38 @@ def replace_hosts_entry(current_smt, new_smt):
...
+    current_smt_ipv4 = current_smt.get_ipv4()
+    current_smt_ipv6 = current_smt.get_ipv6()
+    smt_ipv6_access = has_ipv6_access(new_smt)
+    new_entry = '%s\t' + new_smt.get_FQDN() + '\t' + new_smt.get_name() + '\n'
     for entry in known_hosts:
-        if current_smt_ip in entry:
-            new_hosts += '%s\t%s\t%s\n' % (new_smt.get_ip(),
-                                           new_smt.get_FQDN(),
-                                           new_smt.get_name())
+        if (
```
smarlowucf: This is a bit confusing. Consider using a tiered if else block here instead of multiple if's with continues. Seems like it could be simplified because if smt_ipv6_access then get_ipv6() is always used. And vice versa with ipv4.

regionServiceClient/usr/sbin/registercloudguest
```diff
@@ -254,8 +254,11 @@ if registration_smt:
             region_smt_servers
         )
         if new_target:
+            smt_ip = new_target.get_ipv4()
```
smarlowucf, line 257: Ditto with else clause.

regionServiceClient/usr/sbin/registercloudguest
```diff
@@ -272,16 +275,15 @@ if registration_smt:
 registration_target = None
 tested_smt_servers = []
 for smt in region_smt_servers:
-    smt_ip = smt.get_ip()
-    tested_smt_servers.append(smt_ip)
+    tested_smt_servers.append((smt.get_ipv4(),smt.get_ipv6()))
```
smarlowucf: Space after comma.

regionServiceClient/usr/sbin/registercloudguest
```diff
@@ -402,21 +404,25 @@ elif (os.path.exists(register12) and os.access(register12, os.X_OK)):
...
-                            smt.get_ip() not in failed_smts
+                            smt.get_ipv4() != registration_target.get_ipv4() and
+                            smt.get_ipv4() not in failed_smts
                     ):
                         error_msg = 'Registration with %s failed. Trying %s'
                         logging.error(
                             error_msg % (
-                                registration_target.get_ip(),
-                                smt.get_ip()
+                                str(target_smt_ipv4,target_smt_ipv6),
```
smarlowucf: Missing tuple brackets and space after comma.

smarlowucf (changes requested):

regionServiceClient/lib/cloudregister/registerutils.py
rjschwei: Yes, smt_ipv6 can be None. The trade of is to spend CPU cycles searching for "None" which should never succeed, or spending those cycles in an additional branching instruction with additional book keeping requirements and more code. The current logic is:

if ipv4 or ipv6 and fqdn are in the hosts file all is well

this would have to change into

if ipv4 not found and ipv6 has a value look for ipv6, if ip address entry is found look for fqdn entry

I don't see the benefit of the extra conditional w.r.t. clarity of logic nor expressiveness of the written code.

rjschwei:
@smarlowucf I think I addressed everything, will merge so I can proceed, with the move to SUSE-Enceladus and the removal of SLES 11, so I can get to the stuff I need to do for zypper authentication.

Would appreciate if you could take another look and let me know if you see anything else. There are a number of changes coming in the next week so it'll be easy to include anything I missed here.

Thanks

smarlowucf:
Good, the if/else check could still be updated to be like:

```
if has_ipv6_access(smt_server):
    smt_ip = smt_server.get_ipv6()
else:
    smt_ip = smt_server.get_ipv4()
```

PR #226, opened 2019-03-28
- Code has been moved to SUSE-Enceladus/service-access-config

rjschwei:


rjschwei:
Waited long enough, merging.

